PR #64, opened 2017-05-05
Sphere simulator

mfehr:


mfehr:
Sry, wrong base branch, again...

helenol:
Why not use https://github.com/ethz-asl/voxblox/blob/master/voxblox/include/voxblox/simulation/simulation_world.h ????

PR #66, opened 2017-05-18
Add lambertian color, fix bug with border voxels

helenol:
Bug fixes:
 - Meshing getting invalid/unallocated voxel colors sometimes.
 - getting a block for a 3D point and then a voxel for 3D point would map to invalid voxels

Features:
 - Add a new color mode called lambert_color, which applies lambertian shading (view invariant) to the mesh color. Great for synthetic data.
 - Add a few accessors to TSDF maps.
  - Add a few more options to simulation environment.

@alexmillane @mfehr @ZacharyTaylor

### DIFF
--- a/voxblox/include/voxblox/mesh/mesh_integrator.h
+++ b/voxblox/include/voxblox/mesh/mesh_integrator.h
@@ -276,26 +276,18 @@ class MeshIntegrator {
     for (size_t i = 0; i < mesh->vertices.size(); i++) {
       const Point& vertex = mesh->vertices[i];
       VoxelIndex voxel_index = block.computeVoxelIndexFromCoordinates(vertex);
-      const Point voxel_center =
-          block.computeCoordinatesFromVoxelIndex(voxel_index);
-
-      // Should be within half a voxel of the voxel center in all dimensions, or
-      // it belongs in the other one.
-      const Point dist_from_center = vertex - voxel_center;
-      for (unsigned int j = 0; j < 3; ++j) {
-        if (std::abs(dist_from_center(j)) > voxel_size_ / 2.0) {
-          voxel_index(j) += signum(dist_from_center(j));
-        }
-      }
-
       if (block.isValidVoxelIndex(voxel_index)) {
-        mesh->colors[i] = block.getVoxelByVoxelIndex(voxel_index).color;
+        const TsdfVoxel& voxel = block.getVoxelByVoxelIndex(voxel_index);
+
+        if (voxel.weight >= config_.min_weight) {
+          mesh->colors[i] = voxel.color;
+        }
       } else {
-        // Get the nearest block.
         const Block<TsdfVoxel>::ConstPtr neighbor_block =
             tsdf_layer_->getBlockPtrByCoordinates(vertex);
-        if (neighbor_block) {
-          mesh->colors[i] = neighbor_block->getVoxelByCoordinates(vertex).color;
+        const TsdfVoxel& voxel = neighbor_block->getVoxelByCoordinates(vertex);
+        if (voxel.weight >= config_.min_weight) {
+          mesh->colors[i] = voxel.color;
         }
       }
     }
--- a/voxblox/include/voxblox/simulation/simulation_world.h
+++ b/voxblox/include/voxblox/simulation/simulation_world.h
@@ -49,6 +49,12 @@ class SimulationWorld {
   FloatingPoint getDistanceToPoint(const Point& coords,
                                    FloatingPoint max_dist) const;
 
+  // Set and get the map generation and display bounds.
+  void setBounds(const Point& min_bound, const Point& max_bound) {
+    min_bound_ = min_bound;
+    max_bound_ = max_bound;
+  }
+
   Point getMinBound() const { return min_bound_; }
   Point getMaxBound() const { return max_bound_; }
 
--- a/voxblox/include/voxblox/simulation/simulation_world_inl.h
+++ b/voxblox/include/voxblox/simulation/simulation_world_inl.h
@@ -37,8 +37,13 @@ void SimulationWorld::generateSdfFromWorld(FloatingPoint max_dist,
         layer->allocateBlockPtrByIndex(block_index);
     for (size_t i = 0; i < block->num_voxels(); ++i) {
       VoxelType& voxel = block->getVoxelByLinearIndex(i);
-      Point coords =
-          block->computeCoordinatesFromLinearIndex(i);
+      Point coords = block->computeCoordinatesFromLinearIndex(i);
+      // Check that it's in bounds, otherwise skip it.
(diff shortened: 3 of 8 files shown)


voxblox/include/voxblox/mesh/mesh_integrator.h
```diff
@@ -276,26 +276,18 @@ class MeshIntegrator {
...
       if (block.isValidVoxelIndex(voxel_index)) {
-        mesh->colors[i] = block.getVoxelByVoxelIndex(voxel_index).color;
+        const TsdfVoxel& voxel = block.getVoxelByVoxelIndex(voxel_index);
+
+        if (voxel.weight >= config_.min_weight) {
+          mesh->colors[i] = voxel.color;
+        }
       } else {
-        // Get the nearest block.
         const Block<TsdfVoxel>::ConstPtr neighbor_block =
```
ZacharyTaylor: Now that the offset part is removed isn't this if else statement trying exactly the same thing twice in a row?

voxblox/include/voxblox/simulation/simulation_world.h
```diff
@@ -49,6 +49,12 @@ class SimulationWorld {
   FloatingPoint getDistanceToPoint(const Point& coords,
                                    FloatingPoint max_dist) const;
 
+  // Set and get the map generation and display bounds.
```
ZacharyTaylor: The current defaults for these bounds seem a bit arbitrary, maybe they should be arguments to the initialization?

voxblox/include/voxblox/simulation/simulation_world_inl.h
```diff
@@ -37,8 +37,13 @@ void SimulationWorld::generateSdfFromWorld(FloatingPoint max_dist,
         layer->allocateBlockPtrByIndex(block_index);
     for (size_t i = 0; i < block->num_voxels(); ++i) {
       VoxelType& voxel = block->getVoxelByLinearIndex(i);
-      Point coords =
-          block->computeCoordinatesFromLinearIndex(i);
+      Point coords = block->computeCoordinatesFromLinearIndex(i);
+      // Check that it's in bounds, otherwise skip it.
```
ZacharyTaylor: does it matter if we go 1 voxel out of bounds?

voxblox/src/simulation/simulation_world.cc
```diff
@@ -17,7 +17,19 @@ void SimulationWorld::addGroundLevel(FloatingPoint height) {
...
+      new Plane(Point(x_min, 0.0, 0.0), Point(1.0, 0.0, 0.0)));
+  objects_.emplace_back(
+      new Plane(Point(x_max, 0.0, 0.0), Point(-1.0, 0.0, 0.0)));
+
+  // Y planes:
+  objects_.emplace_back(
+      new Plane(Point(0.0, y_min, 0.0), Point(0.0, 1.0, 0.0)));
+  objects_.emplace_back(
+      new Plane(Point(0.0, y_max, 0.0), Point(0.0, -1.0, 0.0)));
+}
```
ZacharyTaylor: no z boundaries?

voxblox_ros/include/voxblox_ros/mesh_vis.h
```diff
@@ -38,30 +38,44 @@
...
+  kHeight,
+  kNormals,
+  kGray,
+  kLambert,
+  kLambertColor
+};
+
+inline Point lambertShading(const Point& normal, const Point& light,
+                            const Point& color) {
+  return std::max<FloatingPoint>(normal.dot(light), 0.0f) * color;
```
ZacharyTaylor: Can this cause a lot of pure black points?

voxblox_ros/include/voxblox_ros/mesh_vis.h
```diff
@@ -38,30 +38,44 @@
...
 
-inline void lambertColorFromNormal(const Point& normal,
-                            std_msgs::ColorRGBA* color_msg) {
-  static const Point light_dir = Point(0.8f, -0.2f, 0.7f).normalized();
-  static const Point light_dir2 = Point(-0.5f, 0.2f, 0.2f).normalized();
-  static const Point ambient(0.2f, 0.2f, 0.2f);
+inline void lambertColorFromColorAndNormal(const Color& color,
+                                           const Point& normal,
+                                           std_msgs::ColorRGBA* color_msg) {
+  const Point light_dir = Point(0.8f, -0.2f, 0.7f).normalized();
```
ZacharyTaylor: where do these numbers come from?

voxblox_ros/include/voxblox_ros/mesh_vis.h
```diff
@@ -83,8 +97,8 @@ inline void heightColorFromVertex(const Point& vertex,
 }
 
 inline void fillMarkerWithMesh(const MeshLayer::ConstPtr& mesh_layer,
```
alexmillane: Obviously we talked about this already... I have changed this to a constant ref in my branch.

But in general I think we should come to some standard on the use of smart pointers. Obviously they are great, but I think that we should avoid as must as possible passing them around unless they are held for non trivial time. Do you know of a standard on their use?

alexmillane: The reason I say that is that if any client of your function is has a raw pointer to something (rather than a smart pointer) it makes the function unusable, unless you then construct a smart pointer with an empty destructor (hack city). Or maybe I'm wrong and there's something I'm missing.

voxblox/include/voxblox/mesh/mesh_integrator.h
helenol: I think I said this verbally, but there's still the case where a vertex from a certain mesh block actually maps to the adjacent block. This is the case this if statement covers. :)

voxblox/include/voxblox/simulation/simulation_world_inl.h
helenol: Before it was just by-block, and it made the map look sloppy (i.e., much longer on one side depending on where the block boundaries are). I think for visualization purposes it's better to just stick to the actual requested boundaries.

voxblox/src/simulation/simulation_world.cc
helenol: addGroundPlane is for that :)

voxblox_ros/include/voxblox_ros/mesh_vis.h
helenol: Usually not, since ambient light is added to it anyway.

voxblox_ros/include/voxblox_ros/mesh_vis.h
helenol: I think I stole them from openChisel, will add them as parameters.

voxblox_ros/include/voxblox_ros/mesh_vis.h
helenol: I completely agree, can you make a PR for this?

PR #67, opened 2017-05-18
added pointcloud output

ZacharyTaylor:
Outputs the mesh as a pointcloud, the difference between this and the existing surface pointcloud output is that interpolation is used to give the points at the estimated surface location rather then the voxel centers. Makes minimal difference for small voxels. @alexmillane


alexmillane:
LGTM. Great stuff :+1:.

I was also thinking that adding a service to save the pointcloud as ply would be great (similar to what we have in octomap manager). I can do that though, if you like.

ZacharyTaylor:
You could probably just have a service to save the mesh as a ply as all the vertexes are the same. Then there is already a function for it outputMeshLayerAsPly that you would just need to wrap in a service

PR #71, opened 2017-06-08
Added TSDF integration option which better deals with sparse sensor data.

fabianbl:
@helenol could we add this? I used it to integrate sparse landmarks in a separate integrator so far. Maybe there's also a better way to do this.


helenol:
add to whitelist

helenol:
Hey @fabianbl, this is definitely small enough to add as is. :) If there's a lot more changes that need to be done to handle sparsity, I would suggest inheriting a SparseTSDFIntegrator from the current integrator.
But certainly for this small change to weighting, LGTM.

fabianbl:
@helenol Ok, sounds good! Then I will add it like this for the moment and maybe come up with a more sophisticated sparse integrator later :-)

PR #73, opened 2017-06-14
use isfinite

ZacharyTaylor:
isnan will not trigger on some cases where the voxels are invalid (such as -nan) replace with !std::isfinite


ZacharyTaylor:
Actually reading into it more isnan should defiantly pick up -nan, not sure how it was getting through. Anyway I still think using isfinite could catch some more cases

helenol:
LGTM thanks!

PR #75, opened 2017-06-16
Cylinder simulation and more raycasting bugfixes

helenol:
In this PR:
 - Cylinders are my least favorite shape. 
 - Add ray casting and ground truth distance to cylinders in the object simulator.
 - Fix a bug with raycasting that sometimes caused infinite loops (most easy to see with synthetic data).
 - Fix a bug with getGridIndexFromPoint.
 - Slightly refactor ESDF and TSDF servers to just have an updateMesh function rather than timer callback, allowing more easily manually triggering this.

### DIFF
--- a/voxblox/include/voxblox/core/common.h
+++ b/voxblox/include/voxblox/core/common.h
@@ -125,8 +125,10 @@ inline AnyIndex getGridIndexFromPoint(const Point& point,
 // function doesn't always compute the correct grid index for coordinates
 // near the grid cell boundaries.
 inline AnyIndex getGridIndexFromPoint(const Point& scaled_point) {
-  return AnyIndex(std::floor(scaled_point.x()), std::floor(scaled_point.y()),
-                  std::floor(scaled_point.z()));
+  const FloatingPoint kEpsilon = 1e-6;
+  return AnyIndex(std::floor(scaled_point.x() + kEpsilon),
+                  std::floor(scaled_point.y() + kEpsilon),
+                  std::floor(scaled_point.z() + kEpsilon));
 }
 
 inline AnyIndex getGridIndexFromOriginPoint(const Point& point,
--- a/voxblox/include/voxblox/core/common.h
+++ b/voxblox/include/voxblox/core/common.h
@@ -108,6 +108,10 @@ struct Color {
   static const Color Pink() { return Color(255, 0, 127); }
 };
 
+// Constants used across the library.
+constexpr FloatingPoint kEpsilon = 1e-6;  // Used for coordinates.
+constexpr float kFloatEpsilon = 1e-6;     // Used for weights.
+
 // Grid <-> point conversion functions.
 
 // IMPORTANT NOTE: Due the limited accuracy of the FloatingPoint type, this
@@ -115,7 +119,6 @@ struct Color {
 // near the grid cell boundaries.
 inline AnyIndex getGridIndexFromPoint(const Point& point,
                                       const FloatingPoint grid_size_inv) {
-  const FloatingPoint kEpsilon = 1e-6;
   return AnyIndex(std::floor(point.x() * grid_size_inv + kEpsilon),
                   std::floor(point.y() * grid_size_inv + kEpsilon),
                   std::floor(point.z() * grid_size_inv + kEpsilon));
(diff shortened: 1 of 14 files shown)


helenol:
@mfehr

voxblox/include/voxblox/simulation/objects.h
```diff
@@ -242,6 +242,147 @@ class Plane : public Object {
...
+          (point.z() - max_z_limit) * (point.z() - max_z_limit));
+    } else {
+      // Case 3: below cylinder.
+      distance = std::sqrt(
+          std::max((point.head<2>() - center_.head<2>()).squaredNorm() -
+                       radius_ * radius_,
+                   0.0) +
+          (point.z() - min_z_limit) * (point.z() - min_z_limit));
+    }
+    return distance;
```
ZacharyTaylor: I now see why you have such love for cylinders

voxblox/include/voxblox/simulation/objects.h
```diff
@@ -242,6 +242,147 @@ class Plane : public Object {
...
+    FloatingPoint c = vector_E.x() * vector_E.x() +
+                      vector_E.y() * vector_E.y() - radius_ * radius_;
+
+    // t = (-b +- sqrt(b^2 - 4ac))/2a
+    // t only has solutions if b^2 - 4ac >= 0
+    FloatingPoint t1 = -1.0;
+    FloatingPoint t2 = -1.0;
+
+    // Make sure we don't divide by 0.
+    if (std::abs(a) < 1e-6) {
```
ZacharyTaylor: Should we just define a global kEpsilon for voxblox it seems to be needed in a few places

voxblox/src/core/esdf_map.cc
```diff
@@ -19,9 +19,18 @@ bool EsdfMap::getDistanceAndGradientAtPosition(
...
-  bool success = interpolator_.getAdaptiveDistanceAndGradient(
-      position.cast<FloatingPoint>(), &distance_fp, &gradient_fp);
+  bool interpolate = true;
+  bool use_adaptive = false;
+  bool success = false;
+  if (use_adaptive) {
+    success = interpolator_.getAdaptiveDistanceAndGradient(
+        position.cast<FloatingPoint>(), &distance_fp, &gradient_fp);
+  } else {
+    success = interpolator_.getDistance(position.cast<FloatingPoint>(),
```
ZacharyTaylor: why cast here rather then having the right type in the function input?

voxblox/include/voxblox/core/common.h
```diff
@@ -125,8 +125,10 @@ inline AnyIndex getGridIndexFromPoint(const Point& point,
 // function doesn't always compute the correct grid index for coordinates
 // near the grid cell boundaries.
 inline AnyIndex getGridIndexFromPoint(const Point& scaled_point) {
-  return AnyIndex(std::floor(scaled_point.x()), std::floor(scaled_point.y()),
-                  std::floor(scaled_point.z()));
+  const FloatingPoint kEpsilon = 1e-6;
+  return AnyIndex(std::floor(scaled_point.x() + kEpsilon),
+                  std::floor(scaled_point.y() + kEpsilon),
+                  std::floor(scaled_point.z() + kEpsilon));
```
mfehr: @helenol  could you elaborate a bit what exactly the bug was? This doesn't seem quite right...

helenol: @ZacharyTaylor, yes this could happen. However, I think it is much more likely that the number is off by 1e-6 due to compound machine precision problems and actually belongs in the 0 grid, and not the -1 grid where it would be put if it is 1e-6.
@mfehr, the issue here and a bunch of other places was that coordinate -> grid -> coordinate -> grid wouldn't consistently map to the same grid points due to slight machine precision errors, it would get floored to the adjacent grid index. Since we use coordinate to grid conversions heavily for ray casting and many other operations, this was causing infinite loops in ray casting and sometimes accessing unallocated memory. Adding a tiny epsilon fixed this part of the issue in every case I could reproduce.

voxblox/include/voxblox/simulation/objects.h
helenol: Great idea, totally agree.

PR #76, opened 2017-06-20
changed it so occupancy is always advertised

ZacharyTaylor:
occupancy_marker_pub_ is always published to (through a call to publishTsdfOccupiedNodes), but not always advertised, this can cause voxblox to crash with an error about "call to publish() on an invalid publisher".


helenol:
LGTM thanks!!!

PR #77, opened 2017-06-21
Fix/minor compilation warning extinction.

fabianbl:
ptal @helenol @mfehr

### DIFF
--- a/voxblox/include/voxblox/core/block_inl.h
+++ b/voxblox/include/voxblox/core/block_inl.h
@@ -46,7 +46,7 @@ void Block<VoxelType>::getProto(BlockProto* proto) const {
 }
 
 template <typename VoxelType>
-bool Block<VoxelType>::mergeBlock(const Block<VoxelType>& other_block) {
+bool Block<VoxelType>::mergeBlock(const Block<VoxelType>& /*other_block*/) {
   // TODO(mfehr): implement
   LOG(FATAL) << "NOT IMPLEMENTED";
   return false;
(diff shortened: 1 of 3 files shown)


helenol:
LGTM thanks

voxblox/include/voxblox/core/block_inl.h
```diff
@@ -46,9 +46,8 @@ void Block<VoxelType>::getProto(BlockProto* proto) const {
 }
 
 template <typename VoxelType>
-bool Block<VoxelType>::mergeBlock(const Block<VoxelType>& other_block) {
-  // TODO(mfehr): implement
-  LOG(FATAL) << "NOT IMPLEMENTED";
```
mfehr: Thanks for fixing the warning, but why remove this?

voxblox/include/voxblox/core/layer_inl.h
```diff
@@ -165,7 +165,8 @@ bool Layer<VoxelType>::addBlockFromProto(const BlockProto& block_proto,
         if (it == block_map_.end()) {
           block_map_[block_index] = block_ptr;
         } else {
-          it->second->mergeBlock(*block_ptr);
+          CHECK(it->second->mergeBlock(*block_ptr)) << "Block merging not "
```
mfehr: I rather have the crash coming from the function that is not implemented yet. Using the return type as a indicator whether or not the function is implemented is wrong, for some voxel type the return type might be used to communicate some informatio that you can't or shouldn't merge the two blocks.

fabianbl: @mfehr Good point, I changed it back.

mfehr:
@fabianbl thx for fixing this. lgtm

PR #78, opened 2017-06-26
Feature/clearing pointcloud

ZacharyTaylor:
Allows for voxblox to subscribe to points that are not inside an object, but that are also not guaranteed to be on a surface (i.e. they are floating in freespace) . These points will only be used beyond the truncation distance. Not needed for most applications (thus off by default) but could be useful to plan from stereo in a low texture environment. 

Tested on the KITTI dataset with the new version of image_undistort https://github.com/ethz-asl/image_undistort/pull/18

screenshot from 2017-06-26 16-09-10

Left is without freespace points, right is with. Slightly more complete freespace into

### DIFF
--- a/voxblox_ros/src/tsdf_server.cc
+++ b/voxblox_ros/src/tsdf_server.cc
@@ -24,20 +24,28 @@ TsdfServer::TsdfServer(const ros::NodeHandle& nh,
   mesh_pub_ =
       nh_private_.advertise<visualization_msgs::MarkerArray>("mesh", 1, true);
   surface_pointcloud_pub_ =
-      nh_private_.advertise<pcl::PointCloud<pcl::PointXYZRGB> >(
+      nh_private_.advertise<pcl::PointCloud<pcl::PointXYZRGB>>(
           "surface_pointcloud", 1, true);
-  tsdf_pointcloud_pub_ =
-      nh_private_.advertise<pcl::PointCloud<pcl::PointXYZI> >("tsdf_pointcloud",
-                                                              1, true);
+  tsdf_pointcloud_pub_ = nh_private_.advertise<pcl::PointCloud<pcl::PointXYZI>>(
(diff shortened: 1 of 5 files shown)


voxblox_ros/src/tsdf_server.cc
```diff
@@ -157,11 +179,11 @@ void TsdfServer::insertPointcloud(
     // Convert the PCL pointcloud into our awesome format.
     // TODO(helenol): improve...
     // Horrible hack fix to fix color parsing colors in PCL.
-    for (size_t d = 0; d < pointcloud_msg->fields.size(); ++d) {
+    /*for (size_t d = 0; d < pointcloud_msg->fields.size(); ++d) {
```
helenol: Does this need to be removed? This was making many pointclouds not work for me without.

helenol: And if it needs to be removed, please don't just comment it out.

voxblox_ros/src/voxblox_node.cc
```diff
@@ -12,6 +11,7 @@
 #include <std_srvs/Empty.h>
 #include <tf/transform_listener.h>
 #include <visualization_msgs/MarkerArray.h>
+#include <deque>
```
helenol: ? Why move this include?

voxblox_ros/src/tsdf_server.cc
ZacharyTaylor: I actually forgot about this, I had an ugly hack and was going to go back and make it cleaner.

The issue is that I am currently using a boost::bind to pass the freespace flag to the pointcloud callback. However the bind function seems to only work with a callback that has constptr arguments, which prevents us from modifying the incoming message as is done here.

ZacharyTaylor:
Closing as redoing it was easier and quicker than solving all the conflicts. See #111

PR #89, opened 2017-08-02
Feature/fast tsdf

ZacharyTaylor:
Split the Tsdf integrator into a base class which the simple and merged integrator derive from.

Also changed the RayCaster function to its own class and switched it from calculating and returning all points at once to an iterative call.

These changes allowed the introduction of a new "fast" integrator. Each ray is cast backwards from its end towards the camera. If the ray reaches a voxel that has already been updated by another ray, the casting stops and the integerator moves to the next ray. This means that each voxel can only have the information from at most one ray integrated per call. All other information is lost.

When the voxels are sufficiently small, the information loss isn't too bad and the speed gain is massive. Testing on lady and cow I can integrate 1 cm voxels at 10 Hz (though I can't actually visualize it due to poor scaling in some of the meshing and publishing). However the system is limited by memory access issues and is actually slightly slower for large voxels (slower then the merge integrator from about the 10cm voxel mark).

While the integrator is completely thread safe, negotiating memory means the more threads it runs in the slower it goes. I am currently working on a solution to both problems, but if it actually works its a long way off being ready to add it.

### DIFF
--- a/voxblox/include/voxblox/core/layer.h
+++ b/voxblox/include/voxblox/core/layer.h
@@ -199,10 +199,6 @@ class Layer {
   FloatingPoint voxel_size() const { return voxel_size_; }
   size_t voxels_per_side() const { return voxels_per_side_; }
 
-  void rehash(size_t count){
-    block_map_.rehash(count);
-  }
-
   // Serialization tools.
   void getProto(LayerProto* proto) const;
   bool isCompatible(const LayerProto& layer_proto) const;
--- a/voxblox/include/voxblox/core/layer.h
+++ b/voxblox/include/voxblox/core/layer.h
@@ -199,6 +199,10 @@ class Layer {
   FloatingPoint voxel_size() const { return voxel_size_; }
   size_t voxels_per_side() const { return voxels_per_side_; }
 
+  void rehash(size_t count){
+    block_map_.rehash(count);
+  }
+
   // Serialization tools.
   void getProto(LayerProto* proto) const;
   bool isCompatible(const LayerProto& layer_proto) const;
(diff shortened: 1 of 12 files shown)


voxblox/include/voxblox/core/layer.h
```diff
@@ -199,6 +199,10 @@ class Layer {
   FloatingPoint voxel_size() const { return voxel_size_; }
   size_t voxels_per_side() const { return voxels_per_side_; }
 
+  void rehash(size_t count){
```
mfehr: is this needed in this PR?

voxblox/include/voxblox/integrator/integrator_utils.h
```diff
@@ -12,70 +12,89 @@
...
-
-  AnyIndex ray_step_signs(signum(ray_scaled.x()), signum(ray_scaled.y()),
-                          signum(ray_scaled.z()));
-
-  AnyIndex corrected_step(std::max(0, ray_step_signs.x()),
-                          std::max(0, ray_step_signs.y()),
-                          std::max(0, ray_step_signs.z()));
-
-  Point start_scaled_shifted = start_scaled - start_index.cast<FloatingPoint>();
+class RayCaster {
```
ZacharyTaylor: Just tested and seems pretty much the same

Original: 415495 ray casts in 00.749910 seconds
Iterative: 421508 ray casts in 00.746322 seconds

ZacharyTaylor:
@mfehr thanks for the quick and through review. After your comments I went back and had another look at how I was tracking the changed voxels. Managed to get the overhead down quite a bit by switching to a simple set and doing the first check before initializing the raycaster. 

I also removed the leftovers of the thread safety I had in there as it didn't help at all. Its now a bit over twice as fast and strictly faster then merged integration. Would appreciate if you could have another look over the FastTsdfIntegrator::integratePointCloud function before I merge it.

PR #90, opened 2017-08-03
Add linter submodule and move pybind to subfolder

mfehr:
In order to keep our repo nicely formatted and our code clean, I added the linter submodule to check and format new commits.

@ethz-asl/voxbloxers  @ethz-asl/voxblox-students 
Please do the following:
```
cd $VOXBLOX_REPO_ROOT
git pull origin master
git submodule init
git submodule update
./tools/linter/init-git-hooks.py
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,8 @@
+---
+Language: Cpp
+BasedOnStyle: Google
+ColumnLimit: 80
+---
+Language: Proto
+BasedOnStyle: Google
+...
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "pybind11"]
 	path = pybind11
 	url = https://github.com/pybind/pybind11.git
+[submodule "tools/linter"]
+	path = tools/linter
+	url = git@github.com:ethz-asl/linter.git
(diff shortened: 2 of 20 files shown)


mfehr:
@ZacharyTaylor if people hate it, we can easily get rid of it, that's one of the advantages of having it as a submodule.

ZacharyTaylor:
I just tried this today and to get ./tools/linter/init-git-hooks.py to run I first had to go
sudo pip install autopep8 requests
sudo apt install clang-format-3.8

I guess these should be documented as dependencies of voxblox.

I already had clang-format-4.0, does linter have something that is unique to 3.8 or could we change the executable it looks for from "clang-format-diff-3.8" to "clang-format-diff"?

mfehr:
Have you tried this: https://github.com/ethz-asl/linter/blob/master/README.md ?

Hmm yes we should add a pointer to this to the install instructions. However this is only relevant for contributers

mfehr:
I think I had a good reason to use 3.8 but can't remember...  I'm not sure if it's worth supporting multiple versions of clang-format, I think i tried calling the generic clang-format-diff before but something didn't work. But feel free to give it a try.

PR #93, opened 2017-08-04
Change from DCHECK_NOTNULL to DCHECK.

fabianbl:
Would it be ok to have a CHECK instead of a DCHECK there?

### DIFF
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -63,7 +63,8 @@ class RayCaster {
   }
 
   bool nextRayIndex(AnyIndex* ray_index) {
-    *DCHECK_NOTNULL(ray_index) = curr_index_;
+    DCHECK(ray_index != nullptr);
+    *ray_index = curr_index_;
     bool not_at_end = curr_index_ != end_index_;
 
     int t_min_idx;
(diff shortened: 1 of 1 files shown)


ZacharyTaylor:
This function is probably called more times then any other function in voxblox, so if possible I would like to avoid it. What is the motivation for the change, and have you tested it to see what the performance hit is?

fabianbl:
@ZacharyTaylor Ok, if it is called really a lot, performance might suffer from this change. The motivation to change it was that DCHECK_NOTNULL throws an unused-value compiler warning. I now changed it to:

```c++
*DCHECK_NOTNULL(ray_index) = ...
```

Is this ok?

ZacharyTaylor:
Talked to @helenol about this, the issue with the new fix is people tend to ignore lines with DCHECK in it, where as this line is doing something acceptable.

If this is only to get rid of an unused-value compiler warning, can we just change the cmakelist to ignore this warning?

fabianbl:
How do you mean people tend to ignore lines with DCHECK?

Ignoring this warning is possible, but it will not turn the warning off if you are using this header in another repo (e.g. for my case it showed up like 4 times because it's included in different other headers).

What about:
```c++
DCHECK(ray_index != nullptr)
```
?

(actually proposed here: https://github.com/google/glog/blob/master/src/glog/logging.h.in#L1066-L1068 to get of warnings from DCHECK_NOTNULL in release mode)

Issue #92, opened 2017-08-04
Segfaults with MergedTsdfIntegrator.


fabianbl:
Occasionally, I get segfaults after the latest voxblox update.. Any ideas?

```bash
Progress:   0.2% [--------------------------------------------------]I0804 16:13:05.733964 12470 tsdf_integrator.h:312] Went from 67 points to 41 raycasts  and 0 clear rays.
*** Error in `/home/fabian/Code/catkin_ws/devel/lib/voxblox_interface/test_voxblox_interface': double free or corruption (!prev): 0x000000000210e8e0 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7fa767b007e5]
/lib/x86_64-linux-gnu/libc.so.6(+0x8037a)[0x7fa767b0937a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7fa767b0d53c]
/home/fabian/Code/catkin_ws/devel/lib/libvoxblox.so(_ZNSt6vectorIN7voxblox6timing13TimerMapValueESaIS2_EE19_M_emplace_back_auxIJS2_EEEvDpOT_+0xff)[0x7fa76744143f]
/home/fabian/Code/catkin_ws/devel/lib/libvoxblox.so(+0x38e62)[0x7fa767440e62]
/home/fabian/Code/catkin_ws/devel/lib/libvoxblox.so(_ZN7voxblox6timing5TimerC2ERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEb+0x1e)[0x7fa767440fbe]
/home/fabian/Code/catkin_ws/devel/lib/libvoxblox_interface.so(_ZN7voxblox20MergedTsdfIntegrator14integrateVoxelERKN5kindr7minimal26QuatTransformationTemplateIdEERKSt6vectorIN5Eigen6MatrixIdLi3ELi1ELi0ELi3ELi1EEENS8_17aligned_allocatorISA_EEERKS7_INS_5ColorESaISG_EEbbRKSt4pairINS9_IiLi3ELi1ELi0ELi3ELi1EEES7_ImSaImEEERKSt13unordered_mapISM_SO_NS_14BlockIndexHashESt8equal_toISM_ENSB_ISL_IKSM_SO_EEEEPSt5queueINS0_9VoxelInfoESt5dequeIS13_SaIS13_EEE+0x5ef)[0x7fa7699789af]
[...]

PR #95, opened 2017-08-07
Add mutex to timer class.

fabianbl:
Timer class should be thread safe now.

### DIFF
--- a/voxblox/include/voxblox/utils/timing.h
+++ b/voxblox/include/voxblox/utils/timing.h
@@ -24,6 +24,7 @@
 #include <chrono>
 #include <limits>
 #include <map>
+#include <mutex>
 #include <string>
 #include <vector>
 
@@ -176,6 +177,7 @@ class Timing {
   list_t timers_;
   map_t tagMap_;
   size_t maxTagLength_;
+  std::mutex mutex_;
 };
 
 #if ENABLE_MSF_TIMING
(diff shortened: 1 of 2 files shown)


ZacharyTaylor:
The code looks good but I have one major concern with this. If the rest of the function in the thread does not need synchronization then we could get into the situation where the timer is actually forcing the threads to wait and bottlenecking performance.

I would maybe be of the opinion that we either remove any timers from inside concurrent sections or name them via the thread id so they are independent.

mfehr:
I agree with @ZacharyTaylor, if we really want timers inside the threads, then the locking of this implementation is too coarse-grained and would lead to the effects described by Zach.

fabianbl:
It's actually done similarly in maplab, but I can have a look at the performance later. In any case, this might solve ethz-asl/voxblox#92

mfehr:
@ZacharyTaylor I think having separate entries for the thread id would be the right way, but not quite trivial how to implement it, IMO. Should we remove the timers in the concurrent sections for now?

helenol:
I think we should just move timers outside the concurrent sections, which should solve all the problems here?

mfehr:
I agree with @helenol, let's remove them, but then this PR doesn't do any harm. In case we miss a timer in a concurrent section at the very least it's not producing segfaults anymore. So let's remove them and merge this PR, wdyt?

PR #97, opened 2017-08-07
Check for max_ray_length_m independent of allow_clear (fixes #94)

### LINKED ISSUE #94: max_ray_length_m not used properly with all integrators
The `max_ray_length_m` param is used properly with the simple integrator, but the merged integrator ignores it. When for instance using VLP-16 clouds as input this becomes apparent really fast as integration time gets very slow with clouds containing points tens of meters away.

skohlbr:
Without this change, the `max_ray_length_m` parameter is ignored in the merged tsdf integrator when `allow_clear` is `false`.


ffurrer:
Ok to test.

ffurrer:
Ok to test

ffurrer:
Ok to test

helenol:
Thanks for the catch + fix! Merging. :)

PR #100, opened 2017-08-07
A faster prettier voxblox

ZacharyTaylor:
Almost certainly not ready for merging, but I will probably refer to the pr when talking about some of the more questionable hacks in it.

peek 2017-08-07 23-05

Long story short this pull request makes the fast integrator 5x slower, but a hell of a lot prettier. It also threads it and the concurrency is good enough that we are only a bit slower then before.

Meshing is also over 10x faster as we were previously computing unneeded normals.
Most publishing can now be disabled as this bottlenecks high res maps.

Oh my tests we now get a reliable 5Hz on 2cm voxels with voxel carving on.

### DIFF
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -12,9 +12,52 @@
 
 namespace voxblox {
 
-// This class assumes PRE-SCALED coordinates, where one unit = one voxel
-// size. The indices are also returned in this scales coordinate system, which
-// should map to voxel indices.
+// Small class that can be used by multiple threads that need mutually exclusive
+// indexes to the same array, while still covering all elements.
+// The class attempts to make successive indexes far apart so that if the caller
+// also accesses neighboring indexes through thread safe locks, the amount of
+// time a thread must wait on a lock is minimized.
+class ThreadSafeIndex {
+ public:
+  ThreadSafeIndex(const size_t number_of_points, const size_t number_of_threads)
+      : number_of_points_(number_of_points),
+        number_of_threads_(number_of_threads),
+        atomic_idx_(0) {}
+
+  // returns true if index is valid, false otherwise
+  bool getNextIndex(size_t* idx) {
+    DCHECK(idx != nullptr);
+    size_t sequential_idx = atomic_idx_.fetch_add(1);
+
+    if (sequential_idx >= number_of_points_) {
+      return false;
+    } else {
+      *idx = getMixedIndex(sequential_idx);
+      return true;
+    }
+  }
+
+  void reset() { atomic_idx_.store(0); }
+
+ private:
+  size_t getMixedIndex(size_t base_idx) {
+    const size_t number_of_groups = number_of_threads_;
+    const size_t points_per_group = number_of_points_ / number_of_groups;
+
+    const size_t group_num = base_idx % number_of_groups;
+    const size_t position_in_group = base_idx / number_of_groups;
+
+    return group_num * points_per_group + position_in_group;
+  }
+
+  const size_t number_of_points_;
+  const size_t number_of_threads_;
+  std::atomic<size_t> atomic_idx_;
+};
+
+// This class assumes PRE-SCALED coordinates, where one unit = one voxel size.
+// The indices are also returned in this scales coordinate system, which should
+// map to voxel indices.
 class RayCaster {
  public:
   RayCaster(const Point& origin, const Point& point_G,
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -44,6 +44,10 @@ class ThreadSafeIndex {
     const size_t number_of_groups = number_of_threads_;
     const size_t points_per_group = number_of_points_ / number_of_groups;
 
+    if (number_of_groups * points_per_group <= base_idx) {
+      return base_idx;
+    }
+
     const size_t group_num = base_idx % number_of_groups;
     const size_t position_in_group = base_idx / number_of_groups;
 
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -46,17 +46,21 @@ class RayCaster {
     setupRayCaster(start_scaled, end_scaled);
   }
 
+  // returns false if ray terminates at ray_index, true otherwise
   bool nextRayIndex(AnyIndex* ray_index) {
+    if (at_end_) {
+      return false;
+    }
     DCHECK_NOTNULL(ray_index);
     *ray_index = curr_index_;
-    bool not_at_end = curr_index_ != end_index_;
+    at_end_ = curr_index_ == end_index_;
 
     int t_min_idx;
     t_to_next_boundary_.minCoeff(&t_min_idx);
     curr_index_[t_min_idx] += ray_step_signs_[t_min_idx];
     t_to_next_boundary_[t_min_idx] += t_step_size_[t_min_idx];
 
-    return not_at_end;
+    return true;
   }
 
  private:
@@ -65,6 +69,8 @@ class RayCaster {
 
     curr_index_ = getGridIndexFromPoint(start_scaled);
     end_index_ = getGridIndexFromPoint(end_scaled);
+
+    at_end_ = false;
 
     const Ray ray_scaled = end_scaled - start_scaled;
 
@@ -108,6 +114,7 @@ class RayCaster {
   AnyIndex end_index_;
   AnyIndex ray_step_signs_;
   Ray t_step_size_;
+  bool at_end_;
 };
 
 // This function assumes PRE-SCALED coordinates, where one unit = one voxel
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -1,5 +1,5 @@
-#ifndef VOXBLOX_INTEGRATOR_INTEGRATOR_UTILS_H_
-#define VOXBLOX_INTEGRATOR_INTEGRATOR_UTILS_H_
+#ifndef INTEGRATOR_INTEGRATOR_UTILS_H_
+#define INTEGRATOR_INTEGRATOR_UTILS_H_
 
 #include <algorithm>
 #include <vector>
@@ -94,7 +94,7 @@ inline void castRay(
   RayCaster ray_caster(start_scaled, end_scaled);
 
   AnyIndex ray_index;
-  while(ray_caster.nextRayIndex(&ray_index)){
+  while (ray_caster.nextRayIndex(&ray_index)) {
     indices->push_back(ray_index);
   }
 }
@@ -148,4 +148,4 @@ inline void getHierarchicalIndexAlongRay(
 
 }  // namespace voxblox
 
-#endif  // VOXBLOX_INTEGRATOR_INTEGRATOR_UTILS_H_
+#endif  // INTEGRATOR_INTEGRATOR_UTILS_H_
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -51,7 +51,7 @@ class RayCaster {
     if (at_end_) {
       return false;
     }
-    DCHECK_NOTNULL(ray_index);
+    DCHECK(ray_index != nullptr);
     *ray_index = curr_index_;
     at_end_ = curr_index_ == end_index_;
 
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -1,5 +1,5 @@
-#ifndef INTEGRATOR_INTEGRATOR_UTILS_H_
-#define INTEGRATOR_INTEGRATOR_UTILS_H_
+#ifndef VOXBLOX_INTEGRATOR_INTEGRATOR_UTILS_H_
+#define VOXBLOX_INTEGRATOR_INTEGRATOR_UTILS_H_
 
 #include <algorithm>
 #include <vector>
@@ -175,4 +175,4 @@ inline void getHierarchicalIndexAlongRay(
 
 }  // namespace voxblox
 
-#endif  // INTEGRATOR_INTEGRATOR_UTILS_H_
+#endif  // VOXBLOX_INTEGRATOR_INTEGRATOR_UTILS_H_
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -152,7 +152,6 @@ class RayCaster {
                                                             ray_scaled.z());
   }
 
- private:
   Ray t_to_next_boundary_;
   AnyIndex curr_index_;
   AnyIndex end_index_;
--- a/voxblox/include/voxblox/integrator/integrator_utils.h
+++ b/voxblox/include/voxblox/integrator/integrator_utils.h
@@ -17,7 +17,50 @@ namespace voxblox {
 // should map to voxel indices.
 class RayCaster {
  public:
+  RayCaster(const Point& origin, const Point& point_G,
+            const bool is_clearing_ray, const bool voxel_carving_enabled,
+            const FloatingPoint max_ray_length_m,
+            const FloatingPoint voxel_size_inv,
+            const FloatingPoint truncation_distance,
+            const bool cast_from_origin = true) {
+    const Ray unit_ray = (point_G - origin).normalized();
+
+    const Point ray_end = is_clearing_ray
+                              ? origin + unit_ray * max_ray_length_m
+                              : point_G + unit_ray * truncation_distance;
+    const Point ray_start = voxel_carving_enabled
+                                ? origin
+                                : (point_G - unit_ray * truncation_distance);
+
+    const Point start_scaled = ray_start * voxel_size_inv;
+    const Point end_scaled = ray_end * voxel_size_inv;
+
+    if (cast_from_origin) {
+      setupRayCaster(start_scaled, end_scaled);
+    } else {
+      setupRayCaster(end_scaled, start_scaled);
+    }
+  }
+
   RayCaster(const Point& start_scaled, const Point& end_scaled) {
+    setupRayCaster(start_scaled, end_scaled);
+  }
+
+  bool nextRayIndex(AnyIndex* ray_index) {
+    DCHECK_NOTNULL(ray_index);
+    *ray_index = curr_index_;
+    bool not_at_end = curr_index_ != end_index_;
+
+    int t_min_idx;
+    t_to_next_boundary_.minCoeff(&t_min_idx);
+    curr_index_[t_min_idx] += ray_step_signs_[t_min_idx];
+    t_to_next_boundary_[t_min_idx] += t_step_size_[t_min_idx];
+
+    return not_at_end;
+  }
+
+ private:
+  void setupRayCaster(const Point& start_scaled, const Point& end_scaled) {
     constexpr FloatingPoint kTolerance = 1e-6;
 
     curr_index_ = getGridIndexFromPoint(start_scaled);
@@ -29,8 +72,8 @@ class RayCaster {
                                signum(ray_scaled.z()));
 
     const AnyIndex corrected_step(std::max(0, ray_step_signs_.x()),
-                            std::max(0, ray_step_signs_.y()),
-                            std::max(0, ray_step_signs_.z()));
+                                  std::max(0, ray_step_signs_.y()),
+                                  std::max(0, ray_step_signs_.z()));
 
     const Point start_scaled_shifted =
         start_scaled - curr_index_.cast<FloatingPoint>();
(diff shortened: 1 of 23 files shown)


ZacharyTaylor:
The branch is now in a state where it is ready for review.

New in this pull request:
- Ray traces blocks to allow preallocation of all memory needed, allowing concurrent access to the block hash map.
- Uses an array of 4096 mutexes to allow finer grained locking of voxels. This locking happens automatically when updateVoxel is called.
- Bringing together some of the common functions from the merged, fast and simple integrators to reduce code double up.
- Threading of the simple and fast integrator and changing the merging integrator to use the same threading strategy as the others.
- Removed second computation of normals in meshing, as this is done for free in the marching cubes part.
- Added flags for disabling some published information.

helenol:
@ZacharyTaylor when you get the chance, can you merge master into this branch? (Has conflicts)
Will do the review later today.

voxblox/include/voxblox/utils/approx_hash_array.h
```diff
@@ -0,0 +1,153 @@
...
+// only one can be store.
+
+// Basic container, given in an index and get the element that was stored there.
+// There are 2^unmasked_bits_ elements in the container, which element is
+// returned depends on your hashing function.
+// Uses at least 2^unmaksed_bits * sizeof(StoreElement) bytes of ram
+
+namespace voxblox {
+template <size_t unmasked_bits_, typename StoredElement>
+class ApproxHashArray {
```
mfehr: How confident are you that this is all correct? Because for such fundamental containers it would be nice to have at least some basic unit testing :)

voxblox_ros/src/voxblox_node.cc
```diff
@@ -191,15 +197,25 @@ VoxbloxNode::VoxbloxNode(const ros::NodeHandle& nh,
...
   // Advertise topics.
   mesh_pub_ =
       nh_private_.advertise<visualization_msgs::MarkerArray>("mesh", 1, true);
-  surface_pointcloud_pub_ =
-      nh_private_.advertise<pcl::PointCloud<pcl::PointXYZRGB>>(
-          "surface_pointcloud", 1, true);
-  tsdf_pointcloud_pub_ = nh_private_.advertise<pcl::PointCloud<pcl::PointXYZI>>(
-      "tsdf_pointcloud", 1, true);
+
+  if (publish_tsdf_info_) {
```
mfehr: Very nice, when I recently used voxblox I was amazed by how much stuff it publishes :)

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -65,6 +69,159 @@ class TsdfIntegratorBase {
...
+    const Pointcloud& points_C_;
+    const Colors& colors_;
+    const Config& config_;
+    const Point origin_;
+    std::atomic<size_t> atomic_idx_;
+  };
+
+  // Can be used to pre-allocate all the blocks needed by the integrator. This
+  // allows thread-safe lockless block map access
+  void allocateBlocks(const Transformation& T_G_C, const Pointcloud& points_C,
```
helenol: As a small sidenote re: transforming, you do need points_C for the integration as well; not just points_G. While I in general agree with trying not to duplicate work/etc., I'm not 100% sure how huge of a difference at least the pre-transform will make?

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -17,8 +18,11 @@
...
 #include "voxblox/core/voxel.h"
 #include "voxblox/integrator/integrator_utils.h"
+#include "voxblox/utils/approx_hash_array.h"
 #include "voxblox/utils/timing.h"
 
+#include "voxblox/core/block_hash.h"
+
 namespace voxblox {
 
 class TsdfIntegratorBase {
```
ZacharyTaylor: Agreed will split it up

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -65,6 +69,159 @@ class TsdfIntegratorBase {
...
+            Block<TsdfVoxel>::Ptr block =
+                layer_->allocateBlockPtrByIndex(block_idx + AnyIndex(x, y, z));
+            block->updated() = true;
+          }
+        }
+      }
+    }
+  }
+
+  inline TsdfVoxel* getVoxel(const VoxelIndex& global_voxel_idx,
```
helenol: :+1:
Maybe "getVoxelFromGlobalVoxelIndex" as well or whatever

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -65,6 +69,159 @@ class TsdfIntegratorBase {
...
+          points_C_(points_C),
+          colors_(colors),
+          config_(config),
+          origin_(T_G_C.getPosition()),
+          atomic_idx_(0) {
+      DCHECK_EQ(points_C_.size(), colors_.size());
+    }
+
+    bool getNextPoint(size_t* point_idx, Point* point_G, bool* clearing) {
+      while (atomic_idx_.load() < points_C_.size()) {
```
ZacharyTaylor: If the point is too close or too far away it needs to go back and try get the next point, until it hits the end or finds a valid one

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -65,6 +69,159 @@ class TsdfIntegratorBase {
...
+        *point_idx = getMixedIdx(atomic_idx_.fetch_add(1));
```
ZacharyTaylor: true, will change

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -155,109 +316,93 @@ class TsdfIntegratorBase {
...
-      }
+    timing::Timer block_timer("block_allocation");
+    allocateBlocks(T_G_C, points_C, colors);
+    block_timer.Stop();
 
-      FloatingPoint truncation_distance = config_.default_truncation_distance;
+    timing::Timer integrate_timer("integrate");
 
-      const Ray unit_ray = (point_G - origin).normalized();
+    point_getter_ =
```
helenol: Because it needs to know T_G_C, points, etc. on construction and is constructed new every integration request, not once per integrator object?

voxblox/include/voxblox/integrator/tsdf_integrator.h
ZacharyTaylor: I agree the point getter is a bit ugly here. The idea was to place the clearing ray logic inside it so it is all in one place, but I think in practice it made things uglier.

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -65,6 +69,159 @@ class TsdfIntegratorBase {
...
+    std::atomic<size_t> atomic_idx_;
+  };
+
+  // Can be used to pre-allocate all the blocks needed by the integrator. This
+  // allows thread-safe lockless block map access
+  void allocateBlocks(const Transformation& T_G_C, const Pointcloud& points_C,
+                      const Colors& colors) {
+    IndexSet index_set;
+
+    point_getter_ =
```
ZacharyTaylor: was a member so all threads could access it and was shared because it needs parameters on creation. I will redo it as a variable passed to threads.

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -65,6 +69,159 @@ class TsdfIntegratorBase {
...
+    Point point_C, point_G, origin;
+    Color color;
+    bool clearing;
+    while (point_getter_->getNextPoint(&point_C, &point_G, &color, &origin,
+                                       &clearing)) {
+      // immediately check if the voxel the point is in has already been ray
+      // traced (saves time setting up ray tracer for already used points)
+      AnyIndex block_idx =
+          getGridIndexFromPoint(point_G, layer_->block_size_inv());
+      if (index_set.count(block_idx)) {
```
ZacharyTaylor: I believe that this combined with the adding neighbors means that all blocks will be allocated. I need to profile to see how significant the speed gain of this is. If it is not significant I will remove it.

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -155,109 +316,93 @@ class TsdfIntegratorBase {
...
+
     // Pre-compute a list of unique voxels to end on.
     // Create a hashmap: VOXEL INDEX -> index in original cloud.
     BlockHashMapType<std::vector<size_t>>::type voxel_map;
     // This is a hash map (same as above) to all the indices that need to be
     // cleared.
     BlockHashMapType<std::vector<size_t>>::type clear_map;
 
-    bundleRays(T_G_C, points_C, &voxel_map, &clear_map);
+    bundleRays(&voxel_map, &clear_map);
```
helenol: While I have no opinion on whether to refactor PointGetter or not, I agree with Marius on the name obfuscating its purpose. Consider changing it to something that implies it also "stores" the data...

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -406,23 +492,24 @@ class MergedTsdfIntegrator : public TsdfIntegratorBase {
...
-    update_voxels_timer.Stop();
   }
 
   void integrateVoxels(
       const Transformation& T_G_C, const Pointcloud& points_C,
       const Colors& colors, bool enable_anti_grazing, bool clearing_ray,
       const BlockHashMapType<std::vector<size_t>>::type& voxel_map,
       const BlockHashMapType<std::vector<size_t>>::type& clear_map,
-      std::queue<VoxelInfo>* voxel_update_queue, size_t tid) {
+      size_t tid) {
```
helenol: no need for const for fundamental types in my opinion

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -447,127 +534,113 @@ class MergedTsdfIntegrator : public TsdfIntegratorBase {
...
 
     integrate_timer.Stop();
   }
+
+ private:
+  static constexpr FloatingPoint voxel_sub_sample_ = 4.0f;
+  static constexpr size_t max_consecutive_ray_collisions_ = 8;
+  static constexpr size_t masked_bits_ = 20;  // 8 mb of ram per tester
+  static constexpr size_t full_reset_threshold = 10000;
+  ApproxHashSet<masked_bits_, full_reset_threshold> approx_start_tester_;
```
helenol: at least a comment :D

voxblox/include/voxblox/mesh/mesh_integrator.h
```diff
@@ -183,11 +185,7 @@ class MeshIntegrator {
       VoxelIndex corner_index = index + cube_index_offsets_.col(i);
       const VoxelType& voxel = block.getVoxelByVoxelIndex(corner_index);
 
-      // Do not extract a mesh here if one of the corner is unobserved and
-      // outside the truncation region.
-      // TODO(helenol): comment above from open_chisel, but no actual checks
-      // on distance are ever made. Definitely we should skip doing checks of
-      // voxels that are too far from the surface...
+      // Do not extract a mesh here if one of the corner is unobserved.
```
helenol: Cheeeeeers thanks!

voxblox/include/voxblox/utils/approx_hash_array.h
```diff
@@ -0,0 +1,153 @@
...
+// There are several advantages and some very significant disadvantages
+// Advantages-
+// - Simple and blazing fast lockless thread-safe approximate sets
+// - Can be used to provide more fine grain locking of blocks for threading then
+// simply locking the entire layer
+// Disadvantages-
+// - Highly inefficient use of memory (allocates 2^N elements)
+// - Cannot disern between two different elements with the same hash
+// - If the hash of two elements have the same first N elements of their hash,
+// only one can be store.
```
helenol: stored

voxblox/include/voxblox/utils/approx_hash_array.h
```diff
@@ -0,0 +1,153 @@
...
+
+// Basic container, given in an index and get the element that was stored there.
+// There are 2^unmasked_bits_ elements in the container, which element is
+// returned depends on your hashing function.
+// Uses at least 2^unmaksed_bits * sizeof(StoreElement) bytes of ram
```
helenol: masked

voxblox/include/voxblox/utils/approx_hash_array.h
```diff
@@ -0,0 +1,153 @@
...
+  static constexpr size_t pseudo_map_size_ = (1 << unmasked_bits_);
+  static constexpr size_t bit_mask_ = (1 << unmasked_bits_) - 1;
+
+  std::array<StoredElement, pseudo_map_size_> pseudo_map_;
+  BlockIndexHash hasher_;
+};
+
+// Acts as a fast and thread safe set, with the serious limitation of both
+// false-negatives and false positives being possible.
+// Uses at least (2^unmaksed_bits + full_reset_threshold) * sizeof(StoreElement)
```
helenol: masked

voxblox_ros/src/voxblox_node.cc
helenol: 😉

voxblox/include/voxblox/integrator/tsdf_integrator.h
mfehr: @helenol yes we need both points_C and _G, but I think we should not transform them twice if it can be avoided and regarding the PointGetter, I do see how it keeps the integrator functions very short and simple, but IMO they became harder to understand, especially the data flow and it introduced this redundancy regarding the transformation. But let's discuss it when @ZacharyTaylor  is back, unless someone is in a hurry to merge this?

voxblox/include/voxblox/integrator/tsdf_integrator.h
mfehr: @helenol  @ZacharyTaylor yes, I get the no-default-constructor thing, but there is clearly no shared ownership with anyone -> unique_ptr :)

voxblox/include/voxblox/integrator/tsdf_integrator.h
mfehr: the const for fundamental types might not be strictly necessary because the params are a copy anyways, so no harm is done by changing them. However having them const would raise a compilation error should you accidentally assign something to them, which helps prevent bugs. :) (see [here, second answer](https://stackoverflow.com/questions/117293/use-of-const-for-function-parameters))

alexmillane: I agree, I read somewhere once something along the lines of "pass by copy arguments are just local variables declared in a special place, so declare them const as you would any other local variable".

voxblox/include/voxblox/integrator/tsdf_integrator.h
helenol: Ahh yes 👍

ZacharyTaylor:
Ok, I believe I have sorted all the review comments. Its now a lot cleaner and I am quite a bit happier with its layout / performance. @mfehr 

For the block casting for memory pre-allocation I initally ran into issues. Casting every ray was quite expensive (around 100 ms on my dataset) and the fix I had of only casting from unique starting locations and then adding the neighbors of blocks, seemed to have some rare corner cases where it would miss a block.

In the end I swapped to an approach loosely inspired by how infinitam does it. Each thread will try get a voxel from the map, if that block has not been allocated it will instead create a voxel in its own local map and use this instead. Once the integration is done a single thread goes through and adds the contents of all the local maps to the real hashmap. This is slow for the first frame, but otherwise quite efficient (maybe 5-10 ms on same tests).

Other then that changes are:
- Fixed bug where verbose would make mutexes in timers deadlock
- Fixed bug that caused rays to end 1 voxel too soon
- Removed point getter and replaced with a much smaller class that just gets the index, plus an internal function to check if the point is valid.
- Split integrator into a cc file
- Added checks for nullpointers and switched existing DCHECK_NOT_NULLs to DCHECK( != nullptr)
- Threaded mesher
- Added unit tests for approx_hash_array
- Added lots more comments around where the approx_hash_array is used
- Added some benchmarks and images to readme

ZacharyTaylor:
Also @mfehr do you have any idea why it is now not building on ubuntu in jenkins? The differences from where it built correctly are very minor and the error is coming from the tango part,

/usr/bin/ld: CMakeFiles/mesh_tango_tsdf.dir/test/mesh_tango_tsdf.cc.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line

ZacharyTaylor:
test this please

ZacharyTaylor:
test this please

helenol:
FYI this PR fixes:
 - the deadlock caused when printing timings introduced in #95
 - segfaults if running multiple (multi-threaded) TSDF servers at once in the TSDF integrator

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -0,0 +1,567 @@
...
+  // color blending is expensive only do it close to the surface
+  if (std::abs(sdf) < config_.default_truncation_distance) {
+    tsdf_voxel->color = Color::blendTwoColors(
+        tsdf_voxel->color, tsdf_voxel->weight, color, updated_weight);
+  }
+
+  tsdf_voxel->distance =
+      (new_sdf > 0.0) ? std::min(config_.default_truncation_distance, new_sdf)
+                      : std::max(-config_.default_truncation_distance, new_sdf);
+  tsdf_voxel->weight = std::min(config_.max_weight, new_weight);
```
mfehr: how about doing this capping above when you compute the new weight?

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -0,0 +1,567 @@
...
+  if (dist_z > kEpsilon) {
+    return 1.0f / (dist_z * dist_z);
+  }
+  return 0.0f;
+}
+
+void SimpleTsdfIntegrator::integratePointCloud(const Transformation& T_G_C,
+                                               const Pointcloud& points_C,
+                                               const Colors& colors) {
+  std::vector<VoxelMap> temp_voxel_storage(config_.integrator_threads);
```
mfehr: Are you sure this temporary voxel map is worth it? Seems to me there will be a lot of allocating and freeing going on in every integration call.

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -0,0 +1,567 @@
...
+                                               const Pointcloud& points_C,
+                                               const Colors& colors) {
+  timing::Timer integrate_timer("integrate");
+
+  // Pre-compute a list of unique voxels to end on.
+  // Create a hashmap: VOXEL INDEX -> index in original cloud.
+  BlockHashMapType<std::vector<size_t>>::type voxel_map;
+  // This is a hash map (same as above) to all the indices that need to be
+  // cleared.
+  BlockHashMapType<std::vector<size_t>>::type clear_map;
```
mfehr: I guess it does make sense to rename BlockHasMapType, because here it clearly referes to voxel indices, right?

voxblox/test/test_approx_hash_array.cc
```diff
@@ -0,0 +1,119 @@
...
+  // generate 1000 random elements
+  std::vector<AnyIndex> rand_indexes;
+  std::mt19937 gen(1);
+  std::uniform_int_distribution<> dis(1, 1000000);
+  for (int i = 0; i < 1000; ++i) {
+    rand_indexes.push_back(AnyIndex(dis(gen), dis(gen), dis(gen)));
+  }
+
+  // test insert and clearing
+  for (size_t i = 0; i < 50; ++i) {
```
mfehr: 50u

voxblox/src/integrator/tsdf_integrator.cc
ZacharyTaylor: If we want to multi-thread then unless we serialize access to the map, its size cannot change while the threads are working on it. This leaves two choices pre-allocate the blocks before each update or using temporary storage for missed blocks and merging them in later. (Or using locks inside the map but this slows things down a lot).

The pre-allocation of blocks for the cow and lady dataset usually takes 50 to 100 ms per frame. With the hack I was doing of early termination I got this down to 10 to 20 ms, but it still had issues. The cost of copying these missed blocks is usually 5 to 10 ms (much more for the first frame but much less for most of the others). 

It could be improved further, but it doesn't seem to be bottle-necking things.

voxblox/include/voxblox/integrator/integrator_utils.h
```diff
@@ -12,25 +12,121 @@
...
+  }
+
+  // returns false if ray terminates at ray_index, true otherwise
+  bool nextRayIndex(AnyIndex* ray_index) {
+    if (at_end_) {
+      return false;
+    }
+    DCHECK(ray_index != nullptr);
+    *ray_index = curr_index_;
+    at_end_ = curr_index_ == end_index_;
```
ZacharyTaylor: yeah, it was finishing one too early by immediately returning at_end when it should return it on the next call.

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -0,0 +1,567 @@
...
+    return &((*temp_voxel_storage)[global_voxel_idx]);
+  } else {
+    const VoxelIndex local_voxel_idx =
+        getLocalFromGlobalVoxelIndex(global_voxel_idx, voxels_per_side_);
+
+    return &((*last_block)->getVoxelByVoxelIndex(local_voxel_idx));
+  }
+}
+
+// NOT thread safe
```
ZacharyTaylor: If we have a map scope lock then we will need to keep track of how many threads are currently reading or writing to voxels so we can tell when they have all left the map and it is safe to begin operating on it. 

We would also need all these readers to check the lock before moving into the function. This style of reader-writer lock can be difficult to implement without noticeable overhead. C++17 has a std lock type for this, but in C++11 we would need to make our own.

voxblox/src/integrator/tsdf_integrator.cc
ZacharyTaylor: I was mimicking the behavior already present. I am not sure what is best here, will discuss.

voxblox/src/integrator/tsdf_integrator.cc
mfehr: Sure, if it is too much overhead, let's not do it. But just FYI we have a reader-writer lock implementation in aslam_cv2: [reader-writer-lock.h](https://github.com/ethz-asl/aslam_cv2/blob/master/aslam_cv_common/include/aslam/common/reader-writer-lock.h) and [reader-first-reader-writer-lock.h](https://github.com/ethz-asl/aslam_cv2/blob/master/aslam_cv_common/include/aslam/common/reader-first-reader-writer-lock.h)

PR #103, opened 2017-08-28
Fix/brute force eigen alignment fix

ZacharyTaylor:
Voxblox has eigen alignment issues again. They show up when multiple instances of the integrator are run.

At least two classes had issues (CameraModel and RayCaster), but there was probably more. Rather then comb through and fix the issues one by one I am now taking a brute force approach.

This pull request does the following:
* Adds EIGEN_MAKE_ALIGNED_OPERATOR_NEW to every single class and struct
* Creates new aligned types for all the common std containers
* Uses these aligned containers everywhere (with the exception of the serialization and pcl indexes)
* Uses eigens aligned allocator instead of make_shared
* Fixs a large number of small layout issues, as linker picks up a lot when you change most of the files

While blatantly excessive if we make these changes and enforce the use of EIGEN_MAKE_ALIGNED_OPERATOR_NEW and using the aligned containers we should never have to worry about subtle Eigen alignment issues again.

### DIFF
--- a/voxblox/include/voxblox/core/block_hash.h
+++ b/voxblox/include/voxblox/core/block_hash.h
@@ -12,6 +12,8 @@
 namespace voxblox {
 
 struct AnyIndexHash {
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
   static constexpr size_t prime1 = 73856093;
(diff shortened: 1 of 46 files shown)


mfehr:
@ZacharyTaylor awesome :D

PR #105, opened 2017-08-29
Feature/rviz plugin

ZacharyTaylor:
The current marker array we use for meshing needs to send the entire mesh every update. It also sends the color as 4 floats. This combines to mean meshing of even moderately large areas is impossible to do in real time. This plugin makes use of the incremeantal updates to only change the voxels that need it and also transfers only what is needed to do the rendering. This allows much higher render rates and larger renders.

Surfaces seen from the back are also invisible so you can see through ceilings without having to cut them out (originally a bug, but I think it actually improves things).

[example](http://i.imgur.com/aOlBdWw.gifv)


ZacharyTaylor:
Anyone know how to make jenkins rebuild its messages?

voxblox_ros/src/tsdf_server.cc
```diff
@@ -313,11 +310,10 @@ bool TsdfServer::generateMesh() {
   generate_mesh_timer.Stop();
 
   timing::Timer publish_mesh_timer("mesh/publish");
-  visualization_msgs::MarkerArray marker_array;
-  marker_array.markers.resize(1);
-  fillMarkerWithMesh(mesh_layer_, color_mode_, &marker_array.markers[0]);
-  marker_array.markers[0].header.frame_id = world_frame_;
-  mesh_pub_.publish(marker_array);
+  voxblox_msgs::Mesh mesh_msg;
+  fillVoxbloxMeshWithMesh(mesh_layer_, color_mode_, &mesh_msg);
```
mfehr: nit: name is a bit strange, maybe convert/generateToVoxbloxMeshMsg?

voxblox_rviz_plugin/src/voxblox_mesh_visual.cc
```diff
@@ -0,0 +1,74 @@
...
+    } else {
+      std::string object_name = std::to_string(index.x()) + std::string(" ") +
+                                std::to_string(index.y()) + std::string(" ") +
+                                std::to_string(index.z());
+      ogre_object = scene_manager_->createManualObject(object_name);
+      object_map_.insert(std::make_pair(index, ogre_object));
+
+      frame_node_->attachObject(ogre_object);
+    }
+
```
mfehr: check if ogre_object not null?

voxblox_ros/include/voxblox_ros/mesh_vis.h
```diff
@@ -98,6 +99,79 @@ inline void heightColorFromVertex(const Point& vertex,
...
+            break;
+          case kNormals:
+            normalColorFromNormal(mesh->normals[i + j], &color_msg);
+            break;
+          case kLambert:
+            lambertColorFromNormal(mesh->normals[i + j], &color_msg);
+            break;
+          case kLambertColor:
+            lambertColorFromColorAndNormal(mesh->colors[i + j],
+                                           mesh->normals[i], &color_msg);
```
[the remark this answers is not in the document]
ZacharyTaylor: nope, good catch

voxblox_ros/src/tsdf_server.cc
ZacharyTaylor: The other functions that do similar things are all called fill____WithMesh, though I agree it sounds weird. Will rename

PR #107, opened 2017-08-31
improved block allocation logic

ZacharyTaylor:
Changed how blocks not present in the map are handeled.
Previously I created a local map to store the voxels. After the integration was done I then went through and integrated these voxels. This had a very high cost for the first frame and caused noticeable delays with sub-maps.

Now I have a second temporary block map. This map is shared between frames and access is controlled by a mutex. This means in the worst case with an empty layer, all blocks will come from this temporary mutex controlled map and performance will be similar to things running single threaded.

The change also cuts down on copy operations which @mfehr previously noted could be expensive in some cases.

Timings before change:
inserting_missed_voxels     625 00.916614       (00.001467 +- 00.000207)        [00.000671,00.016069]
integrate                   625 12.345203       (00.019752 +- 00.001661)        [00.014607,00.034242]

Timings after change:
inserting_missed_blocks     634 00.017504       (00.000028 +- 00.000005)        [00.000011,00.000094]
integrate                   634 11.797627       (00.018608 +- 00.001380)        [00.013368,00.028746]

### DIFF
--- a/voxblox/include/voxblox/integrator/tsdf_integrator.h
+++ b/voxblox/include/voxblox/integrator/tsdf_integrator.h
@@ -73,15 +73,15 @@ class TsdfIntegratorBase {
   // Will return a pointer to a voxel located at global_voxel_idx in the tsdf
   // layer. Thread safe.
   // Takes in the last_block_idx and last_block to prevent unneeded map lookups.
-  // If the block this voxel would be in has not been allocated, a block in
+  // If this voxel belongs to a block that has not been allocated, a block in
   // temp_block_map_ is created/accessed and a voxel from this map is returned
   // instead. Unlike the layer, accessing temp_block_map_ is controlled via a
   // mutex allowing it to grow during integration.
   // These temporary blocks can be merged into the layer later by calling
   // updateLayerWithStoredBlocks()
-  inline TsdfVoxel* getVoxelPtr(const VoxelIndex& global_voxel_idx,
-                                Block<TsdfVoxel>::Ptr* last_block,
-                                BlockIndex* last_block_idx);
+  inline TsdfVoxel* allocateStorageAndGetVoxelPtr(
+      const VoxelIndex& global_voxel_idx, Block<TsdfVoxel>::Ptr* last_block,
+      BlockIndex* last_block_idx);
 
   // NOT thread safe
   inline void updateLayerWithStoredBlocks();
--- a/voxblox/include/voxblox/integrator/tsdf_integrator.h
+++ b/voxblox/include/voxblox/integrator/tsdf_integrator.h
@@ -33,8 +33,6 @@ class TsdfIntegratorBase {
  public:
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
-  typedef AnyIndexHashMapType<TsdfVoxel>::type VoxelMap;
-
   struct Config {
     EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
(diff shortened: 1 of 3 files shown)


voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -75,16 +73,18 @@ class TsdfIntegratorBase {
...
   // layer. Thread safe.
   // Takes in the last_block_idx and last_block to prevent unneeded map lookups.
-  // If the block this voxel would be in has not been allocated, a voxel in
-  // temp_voxel_storage is allocated and returned instead.
-  // This can be merged into the layer later by calling
-  // updateLayerWithStoredVoxels(temp_voxel_storage)
-  inline TsdfVoxel* findOrTempAllocateVoxelPtr(
-      const VoxelIndex& global_voxel_idx, Block<TsdfVoxel>::Ptr* last_block,
-      BlockIndex* last_block_idx, VoxelMap* temp_voxel_storage) const;
+  // If the block this voxel would be in has not been allocated, a block in
```
mfehr: sth went wrong with this sentence :)

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -75,16 +73,18 @@ class TsdfIntegratorBase {
...
+  // temp_block_map_ is created/accessed and a voxel from this map is returned
+  // instead. Unlike the layer, accessing temp_block_map_ is controlled via a
+  // mutex allowing it to grow during integration.
+  // These temporary blocks can be merged into the layer later by calling
+  // updateLayerWithStoredBlocks()
+  inline TsdfVoxel* getVoxelPtr(const VoxelIndex& global_voxel_idx,
```
mfehr: could you change the name to sth more descriptive, like before? getVoxelPtr sounds like any other getter even though there is some black magic going on.

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -66,67 +67,51 @@ inline TsdfVoxel* TsdfIntegratorBase::findOrTempAllocateVoxelPtr(
   if (block_idx != *last_block_idx) {
     *last_block = layer_->getBlockPtrByIndex(block_idx);
     *last_block_idx = block_idx;
-    if (*last_block != nullptr) {
-      (*last_block)->updated() = true;
-    }
   }
 
   // If no block at this location currently exists, we allocate a temporary
   // voxel that will be merged into the map later
```
mfehr: you now allocate a temporary block, right?

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -66,67 +67,51 @@ inline TsdfVoxel* TsdfIntegratorBase::findOrTempAllocateVoxelPtr(
...
-inline void TsdfIntegratorBase::updateLayerWithStoredVoxels(
-    const VoxelMap& temp_voxel_storage) {
-  BlockIndex last_block_idx;
-  Block<TsdfVoxel>::Ptr block = nullptr;
+    typename Layer<TsdfVoxel>::BlockHashMap::iterator it =
+        temp_block_map_.find(block_idx);
+    if (it != temp_block_map_.end()) {
+      *last_block = it->second;
+    } else {
+      auto insert_status = temp_block_map_.insert(std::make_pair(
```
mfehr: use emplace and drop the make pair?

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -446,40 +418,34 @@ void MergedTsdfIntegrator::integrateRays(
     const AnyIndexHashMapType<AlignedVector<size_t>>::type& clear_map) {
   const Point& origin = T_G_C.getPosition();
 
-  AlignedVector<VoxelMap> temp_voxel_storage(config_.integrator_threads);
-
   // if only 1 thread just do function call, otherwise spawn threads
   if (config_.integrator_threads == 1) {
     integrateVoxels(T_G_C, points_C, colors, enable_anti_grazing, clearing_ray,
-                    voxel_map, clear_map, 0, &(temp_voxel_storage[0]));
+                    voxel_map, clear_map, 0);
```
mfehr: add constexpr for 0

voxblox/src/integrator/tsdf_integrator.cc
ZacharyTaylor: Yes, because of this setting the updated flag has been moved out of the if, as the block will now always exist

ZacharyTaylor:
@mfehr 
The timing of inserting missing is slightly deceptive as some of the work that was done here is now done in the function.

There are two separate mutexes. The first is for accessing the temporary block storage temp_block_mutex_. This one ensures that elements can be added to temp_block_map without any issues.

However, you are right that two threads could have retrieved pointers to the same voxel. That is why before a voxels value is read / changed by the integrator it retrieves the mutex that governs this voxels access and locks it. This happens here https://github.com/ethz-asl/voxblox/blob/master/voxblox/src/integrator/tsdf_integrator.cc#L164

Futher explaination on how we assign mutexes to voxels is given here https://github.com/ethz-asl/voxblox/blob/master/voxblox/include/voxblox/integrator/tsdf_integrator.h#L115

mfehr:
@ZacharyTaylor lgtm, thanks!

ZacharyTaylor:
@mfehr could you approve in the review, I currently can't merge it.

PR #109, opened 2017-08-31
Speed hacks

ZacharyTaylor:
Looked at getting the cpu usage running voxblox down and ended up doing some more optimization.

There are five changes in this pr to speed things up. Let me know if you disagree with these changes (especially 1 and 2 as they could be considered over-optimization)

1) Switched from doubles to floats as the FloatingPoint type. This required changing types in CameraModel

2) When running the fast integrator getLocalFromGlobalVoxelIndex is the third most expensive function in voxblox (1. ApproxHashSet::replaceHash, 2. TsdfIntegratorBase::updateTsdfVoxel). This is due to the modulus operator. I constrained voxels_per_side to be a power of two and replaced this with a bitwise and.

3) Removed the pointer used to index the approx_array and just added an offset. Code is a bit cleaner and a bit faster

4) We were filtering nans twice in pointcloud preprocessing, removed the unneeded one.

5) Add flag to let the fast integrator use the same collision checking for multiple frames. Will give much nosier results but speeds things up if needed (just set to clear checks every frame by default)

Timings on one of the ros bags from jays flight last week:
Before:
inserting_missed_blocks     149  00.014100    
integrate                              149  06.616214   
ptcloud_preprocess             149  00.741457       

After:
inserting_missed_blocks     149  00.012957      
integrate                              149  05.251077       
ptcloud_preprocess             149  00.726723

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,9 @@ A summary of the user setable voxblox_node parameters:
 | `max_weight` | The upper limit for the weight assigned to a voxel | 10000.0 |
 | `use_const_weight` | If true all points along a ray have equal weighting | false |
 | `allow_clear` | If true points beyond the `max_ray_length_m` will be integrated up to this distance | true |
+| `start_voxel_subsampling_factor` | Only used by the fast integrator. Before integration points are inserted into a sub-voxel, only one point is allowed per sub-voxel. This can be thought of as subsampling the pointcloud. The edge length of the sub-voxel is the voxel edge length divided by `start_voxel_subsampling_factor`. | 2 |
+| `max_consecutive_ray_collisions` | Only used by the fast integrator. When a ray is cast by this integrator it detects if any other ray has already passed through the current voxel this scan. If it passes through more than `max_consecutive_ray_collisions` voxels other rays have seen in a row, it is taken to be adding no new information and the casting stops  | 2 |
+| `clear_checks_every_n_frames` | Only used by the fast integrator. Governs how often the sets that indicate if a sub-voxel is full or a voxel has had a ray passed through it are cleared. | 1 |
 | `esdf_max_distance_m` | The maximum distance that the esdf will be calculated out to | 2.0 |
 | `esdf_default_distance_m` | Default distance set for unknown values and values >`esdf_max_distance_m` | 2.0 |
 | `mesh_filename` | Filename output mesh will be saved to, leave blank if no file should be generated | "" |
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ A summary of the user setable voxblox_node parameters:
 | `color_ptcloud_by_weight` | If the pointcloud should be colored by the voxel weighting | false |
 | `method` | Method to use for integrating new readings into the tsdf. Options are "merged", "simple", "merged_discard" and "fast" | "merged" |
 | `tsdf_voxel_size` | The size of the tsdf voxels | 0.2 |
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -178,14 +178,17 @@ A summary of the user setable voxblox_node parameters:
...
-| `tsdf_voxels_per_side` | TSDF voxels per side of an allocated block | 16 |
+| `tsdf_voxels_per_side` | TSDF voxels per side of an allocated block. Must be a power of 2 | 16 |
 | `voxel_carving_enabled` | If true, the entire length of a ray is integrated, if false only the region inside the trunaction distance is used. | true |
 | `truncation_distance` | The truncation distance for the TSDF | 2`tsdf_voxel_size` |
 | `max_ray_length_m` | The maximum range out to which a ray will be cast | 5.0 |
 | `min_ray_length_m` | The point at which the ray casting will start | 0.1 |
 | `max_weight` | The upper limit for the weight assigned to a voxel | 10000.0 |
 | `use_const_weight` | If true all points along a ray have equal weighting | false |
 | `allow_clear` | If true points beyond the `max_ray_length_m` will be integrated up to this distance | true |
+| `start_voxel_subsampling_factor` | Only used by the fast integrator. Before integration points are inserted into a sub-voxel, only one point is allowed per sub-voxel. This can be thought of as subsampling the pointcloud. The edge length of the sub-voxel is the voxel edge length divided by `start_voxel_subsampling_factor`. | 2 |
```
helenol: Nice!!! thanks for adding these.

voxblox/src/utils/camera_model.cc
```diff
@@ -55,31 +53,31 @@ void CameraModel::setIntrinsicsFromFoV(double horizontal_fov,
...
+            -min_distance * tan_half_vertical_fov));
   untransformed_corners_.emplace_back(
-      Eigen::Vector3d(min_distance, -min_distance * tan_half_horizontal_fov,
-                      -min_distance * tan_half_vertical_fov));
+      Point(min_distance, -min_distance * tan_half_horizontal_fov,
+            -min_distance * tan_half_vertical_fov));
   untransformed_corners_.emplace_back(
-      Eigen::Vector3d(min_distance, -min_distance * tan_half_horizontal_fov,
-                      min_distance * tan_half_vertical_fov));
+      Point(min_distance, -min_distance * tan_half_horizontal_fov,
```
helenol: Ugh thanks for cleaning all these up.

voxblox_ros/src/tsdf_server.cc
```diff
@@ -183,15 +191,17 @@ void TsdfServer::insertPointcloud(
 
     timing::Timer ptcloud_timer("ptcloud_preprocess");
 
-    // Filter out NaNs. :|
-    std::vector<int> indices;
-    pcl::removeNaNFromPointCloud(pointcloud_pcl, pointcloud_pcl, indices);
```
helenol: Also nice to swap this out; this function ironically only gets some nans.

voxblox/src/utils/camera_model.cc
ZacharyTaylor: Was actually pretty quick just did a find replace for Eigen::Vector3d and Point

mfehr:
@ZacharyTaylor thanks for yet another speed optimization! However optimization 1 just broke all the interfaces with whoever linked against voxblox... (which we do in maplab)

PR #112, opened 2017-09-02
better naming and dropping normals

ZacharyTaylor:
Ensures multiple instances have unique names in rviz, preventing a crash when using multiple instances of voxblox.
Also took the opportunity to drop the normals from the message, turns out rviz doesn't need them.

### DIFF
--- a/voxblox_rviz_plugin/include/voxblox_rviz_plugin/voxblox_mesh_visual.h
+++ b/voxblox_rviz_plugin/include/voxblox_rviz_plugin/voxblox_mesh_visual.h
@@ -25,8 +25,8 @@ class VoxbloxMeshVisual {
   Ogre::SceneNode* frame_node_;
   Ogre::SceneManager* scene_manager_;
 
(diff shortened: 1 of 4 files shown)


voxblox_rviz_plugin/include/voxblox_rviz_plugin/voxblox_mesh_visual.h
```diff
@@ -25,6 +25,9 @@ class VoxbloxMeshVisual {
   Ogre::SceneNode* frame_node_;
   Ogre::SceneManager* scene_manager_;
 
+  uint instance_number_;
```
helenol: Could we be more explicit here, i.e. uint_64 or uint_32 or uint_8?

helenol: Otherwise I prefer 'unsigned int' over uint.

ZacharyTaylor:
test this please

PR #113, opened 2017-09-04
Feature/esdf speedup

ZacharyTaylor:
@helenol  Had a look at the esdf for a bit. Turns out just implementing the diff param you had in the config makes a huge difference. Even setting it to really small values (current default = 1 mm) makes a large difference.

On the cow and lady dataset with 5 cm voxels:
Before:
esdf                         32 07.620161       (00.238130 +- 00.125271)       
esdf/propagate_tsdf          32 00.200999       (00.006281 +- 00.003060)      
esdf/raise_esdf              32 03.248260       (00.101508 +- 00.058812)      
esdf/update_esdf             32 04.170673       (00.130334 +- 00.064808)       

After:
esdf                         33 01.856199       (00.056248 +- 00.026540)    
esdf/propagate_tsdf          33 00.126689       (00.003839 +- 00.001716)   
esdf/raise_esdf              33 00.320779       (00.009721 +- 00.005105)  
esdf/update_esdf             33 01.408528       (00.042683 +- 00.021910)   

Also had a look at threading but its going to be challenging as the current layout would need a lot of locks and the overhead of them cancels out any benefit. Will leave till after ICRA.

### DIFF
--- a/voxblox/include/voxblox/integrator/esdf_integrator.h
+++ b/voxblox/include/voxblox/integrator/esdf_integrator.h
@@ -151,17 +151,6 @@ class EsdfIntegrator {
   FloatingPoint esdf_voxel_size_;
 
   IndexSet updated_blocks_;
-
-  // We need to prevent simultaneous access to the voxels in the map. We could
-  // put a single mutex on the map or on the blocks, but as voxel updating is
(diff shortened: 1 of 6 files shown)


ZacharyTaylor:
test this please

PR #119, opened 2017-10-16
Clean up ros param handling

mfehr:
I wanted to create a EsdfServer object and configure it using the config structs, but the constructors are all mixing up the config structs with the ros params. I therefore added an option to pass the config structs to the constructors and tried to isolate the ros-params-to-config struct logic a bit. I think it's cleaner this way, if you disagree, ping me and we can discuss it offline.

### DIFF
--- a/voxblox_ros/include/voxblox_ros/ros_params.h
+++ b/voxblox_ros/include/voxblox_ros/ros_params.h
@@ -1,0 +1,127 @@
+#ifndef VOXBLOX_ROS_ROS_PARAMS_H_
+#define VOXBLOX_ROS_ROS_PARAMS_H_
+
+#include <voxblox/core/esdf_map.h>
+#include <voxblox/core/tsdf_map.h>
+#include <voxblox/integrator/esdf_integrator.h>
+#include <voxblox/integrator/tsdf_integrator.h>
+
+namespace voxblox {
+
+inline TsdfMap::Config getTsdfMapConfigFromRosParam(
+    const ros::NodeHandle& nh_private) {
+  TsdfMap::Config tsdf_config;
+
+  // Workaround for OS X on mac mini not having specializations for float
+  // for some reason.
+  double voxel_size = tsdf_config.tsdf_voxel_size;
+  int voxels_per_side = tsdf_config.tsdf_voxels_per_side;
+  nh_private.param("tsdf_voxel_size", voxel_size, voxel_size);
+  nh_private.param("tsdf_voxels_per_side", voxels_per_side, voxels_per_side);
+  if (!isPowerOfTwo(voxels_per_side)) {
+    ROS_ERROR("voxels_per_side must be a power of 2, setting to default value");
+    voxels_per_side = tsdf_config.tsdf_voxels_per_side;
+  }
+
+  tsdf_config.tsdf_voxel_size = static_cast<FloatingPoint>(voxel_size);
+  tsdf_config.tsdf_voxels_per_side = voxels_per_side;
+
+  return tsdf_config;
+}
+
+inline TsdfIntegratorBase::Config getTsdfIntegratorConfigFromRosParam(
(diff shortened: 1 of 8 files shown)


voxblox_ros/include/voxblox_ros/ros_params.h
```diff
@@ -0,0 +1,127 @@
...
+  return tsdf_config;
+}
+
+inline TsdfIntegratorBase::Config getTsdfIntegratorConfigFromRosParam(
+    const ros::NodeHandle& nh_private) {
+  TsdfIntegratorBase::Config integrator_config;
+
+  integrator_config.voxel_carving_enabled = true;
+
+  // Used to be * 4 according to Marius's experience, now * 2.
```
ZacharyTaylor: Can we update the comment, as it currently doesn't match the value.

voxblox_ros/include/voxblox_ros/ros_params.h
```diff
@@ -0,0 +1,127 @@
...
+  return esdf_config;
+}
+
+inline EsdfIntegrator::Config getEsdfIntegratorConfigFromRosParam(
+    const ros::NodeHandle& nh_private) {
+  EsdfIntegrator::Config esdf_integrator_config;
+
+  // TODO(helenol): in the future allow different ESDF and TSDF voxel sizes...
+  // Make sure that this is the same as the truncation distance OR SMALLER!
+  nh_private.param("tsdf_voxel_size", esdf_integrator_config.min_distance_m,
```
ZacharyTaylor: I know, you are just moving these parameters, but is it correct to set the min distance in this way?

voxblox_ros/include/voxblox_ros/ros_params.h
```diff
@@ -0,0 +1,127 @@
...
+                   esdf_integrator_config.min_diff_m);
+
+  return esdf_integrator_config;
+}
+
+}  // namespace voxblox
+
+#endif  // VOXBLOX_ROS_ROS_PARAMS_H_
+
+#include "voxblox_ros/conversions_inl.h"
```
ZacharyTaylor: Why is this include needed?

voxblox_ros/include/voxblox_ros/ros_params.h
mfehr: not sure what the valid message here is... (is it 4 because of my suggestion or because helen made it bigger again because she fixed the weighting? :P)

ZacharyTaylor: Idk, maybe just drop the comment as people can look a few lines down and see its 4x

voxblox_ros/include/voxblox_ros/ros_params.h
mfehr: oops... copy paste error

voxblox_ros/include/voxblox_ros/ros_params.h
mfehr: If I'm not mistaken, this value is used to determine the fixed voxels, the lowest frontier? With this current setting that's roughly a single voxel layer around the surface. Not sure if this is the best solution, but seems ok. @helenol ?

voxblox_ros/include/voxblox_ros/ros_params.h
mfehr: agreed

mfehr:
Note: I'm keeping this branch, do not delete!

voxblox_ros/include/voxblox_ros/ros_params.h
helenol: I agree this is weird... The original thinking was probably a hack workaround for what Marius said (it should be about 1 pixel), but like, this is just a weird way to do it. Will look into fixing this.

helenol:
I agree this is a big improvement! Thanks a lot!

PR #123, opened 2017-10-24
Make SimulationServer class reusable

mfehr:
This PR doesn't change anything in the simulator logic, it's simply a move and slight architecture change. Changes:
 * Cleanup of ros params in simulation server.
 * Moves the simulation server from simulation_eval.cc to a separate header and source file to make it reusable.
 * To use it one now simply implements a derived class and the prepareWorld function, which defines how the world looks like.


voxblox_ros/src/simulation_server.cc
```diff
@@ -0,0 +1,417 @@
...
+                       getEsdfIntegratorConfigFromRosParam(nh_private),
+                       getTsdfMapConfigFromRosParam(nh_private),
+                       getTsdfIntegratorConfigFromRosParam(nh_private)) {}
+
+bool SimulationServer::generatePlausibleViewpoint(FloatingPoint min_distance,
+                                                  Point* ray_origin,
+                                                  Point* ray_direction) const {
+  // Generate a viewpoint at least min_distance from any objects (if you want
+  // just outside an object, just call this with min_distance = 0).
+  constexpr int max_tries = 50;
```
ZacharyTaylor: could we expose this or at least move the 50 out of being hidden in the function?

voxblox_ros/src/simulation_server.cc
```diff
@@ -0,0 +1,417 @@
...
+bool SimulationServer::evaluateVoxel(const TsdfVoxel& voxel_test,
+                                     const TsdfVoxel& voxel_gt,
+                                     FloatingPoint* error) const {
+  if (voxel_test.weight < 1e-6 || voxel_gt.distance < 0.0) {
+    return false;
+  }
+
+  *error = voxel_gt.distance - voxel_test.distance;
+
+  /* if (voxel_gt.distance < -truncation_distance_) {
```
ZacharyTaylor: why is this commented out?

voxblox_ros/src/simulation_server.cc
```diff
@@ -0,0 +1,417 @@
...
+  ROS_INFO_STREAM("TSDF RMSE: " << tsdf_rmse << " ESDF RMSE: " << esdf_rmse);
+
+  ROS_INFO_STREAM("Mesh Timings: " << std::endl << timing::Timing::Print());
+}
+
+void SimulationServer::visualize() {
+  if (!visualize_) {
+    return;
+  }
+  FloatingPoint slice_level = 2.0;
```
ZacharyTaylor: Could this be a parameter, may not always want 2 meters

voxblox_ros/src/simulation_server.cc
helenol: Ahh I think this should be deleted -- this has to do with how behind-surface measurements are penalized in the evaluation, but I eventually settled on only evaluating positive distances (I think this is mentioned in the paper) and the if statement above confirms this. So yeah should just be deleted, here and below.

PR #125, opened 2017-11-13
Allow loading older maps.

helenol:
@mfehr 
This is a bug caused by switching voxblox from double to float -- even though voxels are the same size, they fail a double-float comparison.

Let me know what you think about the extra compatibility messaging; I think it's quite useful since it took me like 20 minutes to figure out what "The layer information read from file is not compatible with the current layer!" meant, but I want to hear what you think about adding extra messaging.

@zarmomin FYI.

### DIFF
--- a/voxblox/include/voxblox/core/layer_inl.h
+++ b/voxblox/include/voxblox/core/layer_inl.h
@@ -216,17 +216,19 @@ bool Layer<VoxelType>::isCompatible(const LayerProto& layer_proto) const {
   compatible &= (getType().compare(layer_proto.type()) == 0);
 
   if (!compatible) {
-    LOG(INFO) << "Voxel size of the loaded map is: " << layer_proto.voxel_size()
-              << " but the current map is: " << voxel_size_ << " check passed? "
-              << (layer_proto.voxel_size() == voxel_size_)
-              << "\nVPS of the loaded map is: " << layer_proto.voxels_per_side()
-              << " but the current map is: " << voxels_per_side_
-              << " check passed? "
-              << (layer_proto.voxels_per_side() == voxels_per_side_)
-              << "\nLayer type of the loaded map is: " << getType()
-              << " but the current map is: " << layer_proto.type()
-              << " check passed? "
-              << (getType().compare(layer_proto.type()) == 0) << std::endl;
+    LOG(WARNING)
+        << "Voxel size of the loaded map is: " << layer_proto.voxel_size()
+        << " but the current map is: " << voxel_size_ << " check passed? "
+        << (std::fabs(layer_proto.voxel_size() - voxel_size_) <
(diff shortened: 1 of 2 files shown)


voxblox/include/voxblox/core/layer_inl.h
```diff
@@ -210,16 +210,32 @@ bool Layer<VoxelType>::addBlockFromProto(const BlockProto& block_proto,
...
 bool Layer<VoxelType>::isCompatible(const LayerProto& layer_proto) const {
   bool compatible = true;
-  compatible &= (layer_proto.voxel_size() == voxel_size_);
+  compatible &= (std::fabs(layer_proto.voxel_size() - voxel_size_) <
+                 std::numeric_limits<FloatingPoint>::epsilon());
   compatible &= (layer_proto.voxels_per_side() == voxels_per_side_);
   compatible &= (getType().compare(layer_proto.type()) == 0);
+
+  if (!compatible) {
+    LOG(INFO) << "Voxel size of the loaded map is: " << layer_proto.voxel_size()
```
mfehr: I would make this a LOG(ERROR) or LOG(WARNING) at the very least. This is not sth that should happen often and LOG(INFO) is easily ignored :)

voxblox/include/voxblox/core/layer_inl.h
```diff
@@ -210,16 +210,32 @@ bool Layer<VoxelType>::addBlockFromProto(const BlockProto& block_proto,
...
+              << (getType().compare(layer_proto.type()) == 0) << std::endl;
+  }
   return compatible;
 }
 
 template <typename VoxelType>
 bool Layer<VoxelType>::isCompatible(const BlockProto& block_proto) const {
   bool compatible = true;
-  compatible &= (block_proto.voxel_size() == voxel_size_);
+  compatible &= (std::fabs(block_proto.voxel_size() - voxel_size_) <
```
mfehr: +1

voxblox_ros/src/tsdf_server.cc
```diff
@@ -80,7 +80,7 @@ TsdfServer::TsdfServer(const ros::NodeHandle& nh,
   tsdf_slice_pub_ = nh_private_.advertise<pcl::PointCloud<pcl::PointXYZI> >(
       "tsdf_slice", 1, true);
 
-  nh_private_.param("pointcloud_queue_size_", pointcloud_queue_size_,
```
mfehr: sry about that, that was probably me when I refactored the ros params...

helenol: 90% chance it was there before the refactor ;)

voxblox/include/voxblox/core/layer_inl.h
helenol: Great, and I'll add a check for double vs float maps into this message as well (since it's trivial).

PR #126, opened 2017-11-15
Add a service call to publish pointclouds on demand.

helenol:
Now you can load maps from file and then still get pointclouds of them, by calling `/voxblox/publish_pointclouds`.

Example surface pointcloud from a pre-recorded map, colored by z-axis:

image

### DIFF
--- a/voxblox_ros/include/voxblox_ros/tsdf_server.h
+++ b/voxblox_ros/include/voxblox_ros/tsdf_server.h
@@ -60,6 +60,7 @@ class TsdfServer {
   virtual void publishSlices();
   virtual void updateMesh();    // Incremental update.
   virtual bool generateMesh();  // Batch update.
+  virtual void publishPointclouds();  // Publishes all available pointclouds.
 
   bool saveMapCallback(voxblox_msgs::FilePath::Request& request,     // NOLINT
                        voxblox_msgs::FilePath::Response& response);  // NOLINT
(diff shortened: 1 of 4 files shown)


voxblox_ros/src/tsdf_server.cc
```diff
@@ -291,6 +293,13 @@ void TsdfServer::publishSlices() {
   tsdf_slice_pub_.publish(pointcloud);
 }
 
+void TsdfServer::publishPointclouds() {
+  // Combined function to publish everything we got.
```
ffurrer: Not sure if this is the best method name for its purpose.

voxblox_ros/include/voxblox_ros/tsdf_server.h
```diff
@@ -60,13 +60,17 @@ class TsdfServer {
...
+  virtual void publishPointclouds();  // Publishes all available pointclouds.
 
   bool saveMapCallback(voxblox_msgs::FilePath::Request& request,     // NOLINT
                        voxblox_msgs::FilePath::Response& response);  // NOLINT
   bool loadMapCallback(voxblox_msgs::FilePath::Request& request,     // NOLINT
                        voxblox_msgs::FilePath::Response& response);  // NOLINT
   bool generateMeshCallback(std_srvs::Empty::Request& request,       // NOLINT
                             std_srvs::Empty::Response& response);    // NOLINT
+  bool publishPointcloudsCallback(
+      std_srvs::Empty::Request& request,     // NOLINT
```
ffurrer: const request? here, and all the requests in the service callbacks?

helenol: Doesn't match the file pattern for advertising services (i.e., won't compile with const there). Is there another way to do this? I haven't found one :/

voxblox_ros/src/tsdf_server.cc
helenol: "Everything we got" which is a pointcloud :) So all the stuff published by this function is a pointcloud message; some are surface pointclouds and some are pointclouds with all the points in the map.

PR #130, opened 2017-11-21
made methods protected not private and un-inlined methods implemented…

ffurrer:
… in cc files

The inlined functions could also be placed in the header files, but like it was before, linking to the inlined methods failed in different packages.

So, if this would drastically decrease speed we can move the inlined methods implementations to the headers.

### DIFF
--- a/voxblox/include/voxblox/integrator/tsdf_integrator.h
+++ b/voxblox/include/voxblox/integrator/tsdf_integrator.h
@@ -83,15 +83,15 @@ class TsdfIntegratorBase {
   // mutex allowing it to grow during integration.
   // These temporary blocks can be merged into the layer later by calling
   // updateLayerWithStoredBlocks()
-  inline TsdfVoxel* allocateStorageAndGetVoxelPtr(
+  TsdfVoxel* allocateStorageAndGetVoxelPtr(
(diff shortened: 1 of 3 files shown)


voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -83,15 +83,15 @@ class TsdfIntegratorBase {
   // mutex allowing it to grow during integration.
   // These temporary blocks can be merged into the layer later by calling
   // updateLayerWithStoredBlocks()
-  inline TsdfVoxel* allocateStorageAndGetVoxelPtr(
+  TsdfVoxel* allocateStorageAndGetVoxelPtr(
```
ZacharyTaylor: This function and updateTsdfVoxel are probably the two most called functions in voxblox. Though I suspect that even without the inline specification that the compiler would usually inline them anyway. 

Maybe just check the timings before and after the change but I am guessing this won't make any difference to run time.

margaritaG: I checked the timings of this branch vs current voxblox master using the voxblox_ros cow_and_lady_dataset.launch and can't see any difference between the run times of the two versions.

ffurrer: @margaritaG thanks, will merge this then :)

PR #133, opened 2017-11-22
Feature/add function to center layer

mfehr:
Changes:
 * Adds a function to center the grid, i.e. modify the block indices such that the grid origin will become the centroid of all allocated blocks.
 * Cleanup the test layer function to remove the code triplication.
 * Add unit test for the new function

### DIFF
--- a/voxblox/CMakeLists.txt
+++ b/voxblox/CMakeLists.txt
@@ -56,6 +56,7 @@ set("${PROJECT_NAME}_SRCS"
   src/mesh/marching_cubes.cc
   src/simulation/objects.cc
   src/simulation/simulation_world.cc
+  src/test/layer_test_utils.cc
   src/utils/camera_model.cc
   src/utils/layer_utils.cc
   src/utils/protobuf_utils.cc
--- a/voxblox/CMakeLists.txt
+++ b/voxblox/CMakeLists.txt
@@ -166,6 +166,11 @@ catkin_add_gtest(test_layer
 )
 target_link_libraries(test_layer ${PROJECT_NAME} ${catkin_LIBRARIES})
 
+catkin_add_gtest(test_layer_utils
+  test/test_layer_utils.cc
+)
+target_link_libraries(test_layer_utils ${PROJECT_NAME} ${catkin_LIBRARIES})
+
 ##########
 # EXPORT #
 ##########
--- a/voxblox/include/voxblox/utils/layer_utils.h
+++ b/voxblox/include/voxblox/utils/layer_utils.h
@@ -1,6 +1,11 @@
-#ifndef VOXBLOX_TEST_LAYER_UTILS_H_
-#define VOXBLOX_TEST_LAYER_UTILS_H_
+#ifndef VOXBLOX_UTILS_LAYER_UTILS_H_
+#define VOXBLOX_UTILS_LAYER_UTILS_H_
 
+#include <utility>
+
+#include <Eigen/Core>
+
+#include "voxblox/core/common.h"
 #include "voxblox/core/layer.h"
 #include "voxblox/core/voxel.h"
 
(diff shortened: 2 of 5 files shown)


voxblox/CMakeLists.txt
```diff
@@ -165,6 +166,11 @@ catkin_add_gtest(test_layer
 )
 target_link_libraries(test_layer ${PROJECT_NAME} ${catkin_LIBRARIES})
 
+catkin_add_gtest(test_layer_utils
+  test/test_layer_utils.cc
+)
+target_link_libraries(test_layer_utils ${PROJECT_NAME} ${catkin_LIBRARIES})
```
ffurrer: I don't think you actually need ${catkin_LIBRARIES} here, should be done by catkin simple.

voxblox/include/voxblox/utils/layer_utils.h
```diff
@@ -111,7 +116,61 @@ bool isSameVoxel(const EsdfVoxel& voxel_A, const EsdfVoxel& voxel_B);
...
+    centroid += centroid;
+  }
+  centroid /= static_cast<FloatingPoint>(block_indices.size());
+
+  // Round to nearest block index to centroid.
+  const BlockIndex index_centroid =
+      (centroid + 0.5 * Point::Ones()).cast<IndexElement>();
+
+  LOG(ERROR) << "The index centroid of all allocated blocks is: "
+             << index_centroid;
```
ffurrer: rm?

voxblox/include/voxblox/utils/layer_utils.h
```diff
@@ -111,7 +116,61 @@ bool isSameVoxel(const EsdfVoxel& voxel_A, const EsdfVoxel& voxel_B);
...
+
+  // Loop over all blocks and change their spatial indices.
+  // The only way to do this is to remove them all, store them in a temporary
+  // hash map and re-insert them again. This sounds worse than it is, the blocks
+  // are all shared ptrs and therefore only a negligable amount of real memory
```
ffurrer: nit: negligible

voxblox/CMakeLists.txt
helenol: Hmm but add gtest isn't a cs command? Not sure on this...

voxblox/include/voxblox/utils/layer_utils.h
```diff
@@ -111,7 +116,61 @@ bool isSameVoxel(const EsdfVoxel& voxel_A, const EsdfVoxel& voxel_B);
...
+  // operations is necessary.
+  const size_t num_allocated_blocks_before =
+      layer->getNumberOfAllocatedBlocks();
+  typename Layer<VoxelType>::BlockHashMap temporary_map;
+  layer->getAllAllocatedBlocks(&block_indices);
+  for (const BlockIndex block_index : block_indices) {
```
helenol: const ref?

voxblox/include/voxblox/utils/layer_utils.h
```diff
@@ -111,7 +116,61 @@ bool isSameVoxel(const EsdfVoxel& voxel_A, const EsdfVoxel& voxel_B);
...
+void centerBlocksOfLayer(Layer<VoxelType>* layer, Point* new_layer_origin) {
+  CHECK_NOTNULL(layer);
+  CHECK_NOTNULL(new_layer_origin);
+
+  // Compute the exact cenroid of all allocated block indices.
+  Point centroid = Point::Zero();
+  BlockIndexList block_indices;
+  layer->getAllAllocatedBlocks(&block_indices);
+  for (const BlockIndex block_index : block_indices) {
+    centroid += centroid;
```
ZacharyTaylor: wait am I missing something or is centroid still = Point::Zero() at this stage, making this do nothing?

mfehr: hahaha I think you are right... This function currently doens't do anything. Good catch.  I guess I was just too tired :P

voxblox/CMakeLists.txt
mfehr: I guess this is a more general discussion, not limited to this PR :)
https://github.com/ethz-asl/voxblox/issues/134

voxblox/include/voxblox/utils/layer_utils.h
mfehr: I can't believe how I could fuck this up even with a unit test.... I just set the expected returned offset 
 values to 0 to see what the output would be the first time I run it. However the function returned 0 offset and the unit test passed and I probably got distracted in the meantime, so when I continued it all seemed in order -> Done :D

PR #139, opened 2017-12-01
Fix/compiler warnings

cwitting:
@helenol @mfehr So voxblox compiles without warnings using -Wall (only some weird catkin warning). However all these cast are probably only symptom handling and not the right way to do it. 
But the other fixes should be legit.

### DIFF
--- a/voxblox/CMakeLists.txt
+++ b/voxblox/CMakeLists.txt
@@ -5,7 +5,7 @@ find_package(catkin_simple REQUIRED)
 catkin_simple()
 
 set(CMAKE_MACOSX_RPATH 0)
-add_definitions(-std=c++11 -Wno-sign-compare -Wno-unused-value)
+add_definitions(-std=c++11 -Wall)
 
 if (EXISTS "${CMAKE_CURRENT_LIST_DIR}/../tools/pybind11/CMakeLists.txt")
   set(HAVE_PYBIND11 TRUE)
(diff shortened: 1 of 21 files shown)


helenol:
add to whitelist

ZacharyTaylor:
Thanks for going through and sorting these warnings. One issue though is that we usually try to stick to the google style guide and the use of c style castings goes against this https://google.github.io/styleguide/cppguide.html#Casting could you swap them to static_cast ?

voxblox/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ find_package(catkin_simple REQUIRED)
 catkin_simple()
 
 set(CMAKE_MACOSX_RPATH 0)
-add_definitions(-std=c++11 -Wno-sign-compare -Wno-unused-value)
+add_definitions(-std=c++11 -Wall)
```
mfehr: Awesome :+1:

voxblox/src/core/esdf_map.cc
```diff
@@ -17,7 +17,7 @@ bool EsdfMap::getDistanceAtPosition(const Eigen::Vector3d& position,
 bool EsdfMap::getDistanceAndGradientAtPosition(
     const Eigen::Vector3d& position, double* distance,
     Eigen::Vector3d* gradient) const {
-  FloatingPoint distance_fp;
+  FloatingPoint distance_fp = 0;
```
mfehr: 0.0

PR #141, opened 2017-12-01
Fix warnings raised by -Wall

mfehr:
We use -Wextra in maplab, therefore voxblox was still raising warnings.
This should fix all of them.

### DIFF
--- a/voxblox/include/voxblox/integrator/esdf_occ_integrator.h
+++ b/voxblox/include/voxblox/integrator/esdf_occ_integrator.h
@@ -38,8 +38,7 @@ class EsdfOccIntegrator {
   // Fixed is overloaded as occupied in this case.
   // Only batch operations are currently supported for the occupancy map.
   void updateFromOccLayerBatch();
-  void updateFromOccBlocks(const BlockIndexList& occ_blocks,
-                           bool push_neighbors);
+  void updateFromOccBlocks(const BlockIndexList& occ_blocks);
 
   void processOpenSet();
 
--- a/voxblox/include/voxblox/integrator/merge_integration.h
+++ b/voxblox/include/voxblox/integrator/merge_integration.h
@@ -39,7 +39,8 @@ void mergeBlockAIntoBlockB(const Block<VoxelType>& block_A,
     block_B->has_data() = true;
     block_B->updated() = true;
 
-    for (IndexElement voxel_idx = 0; voxel_idx < block_B->num_voxels();
+    for (IndexElement voxel_idx = 0;
+         voxel_idx < static_cast<IndexElement>(block_B->num_voxels());
          ++voxel_idx) {
       mergeVoxelAIntoVoxelB<VoxelType>(
           block_A.getVoxelByLinearIndex(voxel_idx),
@@ -221,7 +222,7 @@ void transformLayer(const Layer<VoxelType>& layer_in,
         layer_out->allocateBlockPtrByIndex(block_idx);
 
     for (IndexElement voxel_idx = 0;
-          voxel_idx < static_cast<IndexElement>(block->num_voxels());
+         voxel_idx < static_cast<IndexElement>(block->num_voxels());
          ++voxel_idx) {
       VoxelType& voxel = block->getVoxelByLinearIndex(voxel_idx);
 
--- a/voxblox/include/voxblox/integrator/tsdf_integrator.h
+++ b/voxblox/include/voxblox/integrator/tsdf_integrator.h
@@ -83,18 +83,17 @@ class TsdfIntegratorBase {
   // mutex allowing it to grow during integration.
   // These temporary blocks can be merged into the layer later by calling
   // updateLayerWithStoredBlocks()
-  TsdfVoxel* allocateStorageAndGetVoxelPtr(
-      const VoxelIndex& global_voxel_idx, Block<TsdfVoxel>::Ptr* last_block,
-      BlockIndex* last_block_idx);
+  TsdfVoxel* allocateStorageAndGetVoxelPtr(const VoxelIndex& global_voxel_idx,
+                                           Block<TsdfVoxel>::Ptr* last_block,
+                                           BlockIndex* last_block_idx);
 
   // NOT thread safe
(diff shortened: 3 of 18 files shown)


voxblox/include/voxblox/integrator/esdf_occ_integrator.h
```diff
@@ -38,8 +38,7 @@ class EsdfOccIntegrator {
   // Fixed is overloaded as occupied in this case.
   // Only batch operations are currently supported for the occupancy map.
   void updateFromOccLayerBatch();
-  void updateFromOccBlocks(const BlockIndexList& occ_blocks,
-                           bool push_neighbors);
```
mfehr: @helenol this push_neighbor was never used so I removed it, let me know if this is ok or if it should have been used and this was a bug.

voxblox/include/voxblox/integrator/merge_integration.h
```diff
@@ -221,7 +222,7 @@ void transformLayer(const Layer<VoxelType>& layer_in,
         layer_out->allocateBlockPtrByIndex(block_idx);
 
     for (IndexElement voxel_idx = 0;
-          voxel_idx < static_cast<IndexElement>(block->num_voxels());
+         voxel_idx < static_cast<IndexElement>(block->num_voxels());
```
mfehr: this was already fixe by @cwitting  but when I resolved a merge conflict in this file in another PR it must have undone his changes.

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -163,12 +162,10 @@ class MergedTsdfIntegrator : public TsdfIntegratorBase {
                            const bool freespace_points = false);
 
  protected:
-  void bundleRays(
-      const Transformation& T_G_C, const Pointcloud& points_C,
-      const Colors& colors, const bool freespace_points,
-      ThreadSafeIndex* index_getter,
-      AnyIndexHashMapType<AlignedVector<size_t>>::type* voxel_map,
-      AnyIndexHashMapType<AlignedVector<size_t>>::type* clear_map);
+  void bundleRays(const Transformation& T_G_C, const Pointcloud& points_C,
```
mfehr: here I removed the color argument, since it was not required inside.

voxblox/include/voxblox/simulation/objects.h
```diff
@@ -27,19 +27,16 @@ class Object {
   virtual ~Object() {}
 
   // Map-building accessors.
-  virtual FloatingPoint getDistanceToPoint(const Point& point) const {
-    return 0.0;
-  }
+  virtual FloatingPoint getDistanceToPoint(const Point& point) const = 0;
```
mfehr: @helenol I removed the dummy implementations, this will force the descendants to implement these functions, which is what you want I guess.

voxblox/include/voxblox/simulation/objects.h
```diff
@@ -142,7 +139,7 @@ class Cube : public Object {
 
   virtual bool getRayIntersection(const Point& ray_origin,
                                   const Point& ray_direction,
-                                  FloatingPoint max_dist,
+                                  FloatingPoint /*max_dist*/,
```
mfehr: @helenol  this could be a bug, in the other objects you used this max_dist parameter, but here it's not used. wdyt?

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -522,7 +521,7 @@ void FastTsdfIntegrator::integratePointCloud(const Transformation& T_G_C,
 
   integration_start_time_ = std::chrono::steady_clock::now();
 
-  static long int reset_counter = 0;
+  static int64_t reset_counter = 0;
```
mfehr: @cwitting the linter was complaining about the C-type ling int.

voxblox_ros/src/transformer.cc
```diff
@@ -112,9 +112,7 @@ bool Transformer::lookupTransformTf(const std::string& from_frame,
   return true;
 }
 
-bool Transformer::lookupTransformQueue(const std::string& from_frame,
-                                       const std::string& to_frame,
-                                       const ros::Time& timestamp,
+bool Transformer::lookupTransformQueue(const ros::Time& timestamp,
```
mfehr: @helenol  this function (which has a duplicate in voxblox node) didn't use the from/to frame names, I guess that was on purpose?

voxblox/include/voxblox/integrator/esdf_occ_integrator.h
helenol: I think no problem to remove, probably meant to implement it at some point but clearly didn't.

voxblox/include/voxblox/simulation/objects.h
helenol: 👍

voxblox/include/voxblox/simulation/objects.h
helenol: Yeah this is a bug. Should add one last check for the distance before returning. THANKS!!!!!!
```
     if (t > max_dist) {
       return false;
     }
```

voxblox_ros/src/transformer.cc
helenol: Just matching the same interface as the other function.... MIght make sense to keep it that way in case of the future if we actually do smarter resolution of queues? But this is also frickin' unlikely to ever happen to be honest. What do you think?

mfehr: I think we can always extend it again if this ever happens. Keeping them will just confuse people, IMO.

voxblox/include/voxblox/simulation/objects.h
```diff
@@ -188,6 +185,11 @@ class Cube : public Object {
         return false;
       }
     }
+
+    if (t > max_dist) {
+      return false;
+    }
+
```
mfehr: @helenol is this correct?

helenol: Si Señor

mfehr:
@helenol thanks for the immediate review

PR #143, opened 2017-12-13
Adds functionality to publish the tsdf layer as well upon service call

zarmomin:


helenol:
add to whitelist

voxblox_ros/src/tsdf_server.cc
```diff
@@ -302,12 +302,25 @@ void TsdfServer::publishSlices() {
...
+  }
+}
+
 void TsdfServer::publishPointclouds() {
   // Combined function to publish all possible pointcloud messages -- surface
   // pointclouds, updated points, and occupied points.
   publishAllUpdatedTsdfVoxels();
   publishTsdfSurfacePoints();
   publishTsdfOccupiedNodes();
+  publishMap();
```
helenol: So right now the TSDF map gets published on mesh updates periodically, if these are turned on.
Instead of lumping this with publishPointclouds, can we add a new service call called publish_map or publish_layers? I think this would make a bit more sense. :)

zarmomin: I added a service call to "publish_map" but left the publishing-event in updateMesh() https://github.com/ethz-asl/voxblox/blob/4be2a2f051fcc93b1463e8b3e690dc41291a9949/voxblox_ros/src/tsdf_server.cc#L318
in place

PR #147, opened 2018-01-09
don't do clearing without carving

ZacharyTaylor:
Previously if clearing was enabled and carving was disabled the system would begin clearing backwards beyond the max ray distance. This fix will just gives a clearing ray of length 0 instead.
@mfehr @margaritaG 
https://github.com/ethz-asl/voxblox_gsm/issues/12


ZacharyTaylor:
test this please

PR #148, opened 2018-01-15
Add support for concatenating multiple layers when serializing.

helenol:
Now you can serialize multiple layers on the same file, if you turn multiple layer support on when reading back. Default behavior remains the same.
Also split up the io functionality a bit and added comments about what the different functions do, since it took me ages to figure out...

### DIFF
--- a/voxblox/include/voxblox/core/layer.h
+++ b/voxblox/include/voxblox/core/layer.h
@@ -224,15 +224,11 @@ class Layer {
   void getProto(LayerProto* proto) const;
   bool isCompatible(const LayerProto& layer_proto) const;
   bool isCompatible(const BlockProto& layer_proto) const;
-  bool saveToFile(const std::string& file_path) const;
-  bool saveToFile(const std::string& file_path, bool clear_file) const;
+  bool saveToFile(const std::string& file_path, bool clear_file = true) const;
   // Default behavior is to clear the file.
   bool saveSubsetToFile(const std::string& file_path,
                         BlockIndexList blocks_to_include,
-                        bool include_all_blocks) const;
-  bool saveSubsetToFile(const std::string& file_path,
-                        BlockIndexList blocks_to_include,
-                        bool include_all_blocks, bool clear_file) const;
+                        bool include_all_blocks, bool clear_file = true) const;
   bool addBlockFromProto(const BlockProto& block_proto,
                          BlockMergingStrategy strategy);
 
--- a/voxblox/include/voxblox/core/layer.h
+++ b/voxblox/include/voxblox/core/layer.h
@@ -127,8 +127,8 @@ class Layer {
                    voxels_per_side_, voxel_size_,
                    getOriginPointFromGridIndex(index, block_size_)));
 
-    DCHECK(insert_status.second)
-        << "Block already exists when allocating at " << index.transpose();
+    DCHECK(insert_status.second) << "Block already exists when allocating at "
+                                 << index.transpose();
 
     DCHECK(insert_status.first->second);
     DCHECK_EQ(insert_status.first->first, index);
@@ -225,9 +225,14 @@ class Layer {
   bool isCompatible(const LayerProto& layer_proto) const;
   bool isCompatible(const BlockProto& layer_proto) const;
   bool saveToFile(const std::string& file_path) const;
+  bool saveToFile(const std::string& file_path, bool clear_file) const;
+  // Default behavior is to clear the file.
   bool saveSubsetToFile(const std::string& file_path,
                         BlockIndexList blocks_to_include,
                         bool include_all_blocks) const;
+  bool saveSubsetToFile(const std::string& file_path,
+                        BlockIndexList blocks_to_include,
+                        bool include_all_blocks, bool clear_file) const;
   bool addBlockFromProto(const BlockProto& block_proto,
                          BlockMergingStrategy strategy);
 
--- a/voxblox/include/voxblox/io/layer_io_inl.h
+++ b/voxblox/include/voxblox/io/layer_io_inl.h
@@ -105,13 +105,13 @@ bool LoadBlocksFromFile(
     const std::string& file_path,
     typename Layer<VoxelType>::BlockMergingStrategy strategy,
     Layer<VoxelType>* layer_ptr) {
-  bool multiple_layer_support = false;
+  constexpr bool multiple_layer_support = false;
   return LoadBlocksFromFile(file_path, strategy, multiple_layer_support,
                             layer_ptr);
 }
 
 template <typename VoxelType>
-bool LoadLayer(const std::string& file_path, bool multiple_layer_support,
+bool LoadLayer(const std::string& file_path, const bool multiple_layer_support,
                typename Layer<VoxelType>::Ptr* layer_ptr) {
   CHECK_NOTNULL(layer_ptr);
   CHECK(!file_path.empty());
@@ -229,7 +229,7 @@ bool SaveLayer(const Layer<VoxelType>& layer, const std::string& file_path,
 template <typename VoxelType>
 bool SaveLayerSubset(const Layer<VoxelType>& layer,
                      const std::string& file_path,
-                     BlockIndexList blocks_to_include,
+                     const BlockIndexList& blocks_to_include,
                      bool include_all_blocks) {
   CHECK(!file_path.empty());
   return layer.saveSubsetToFile(file_path, blocks_to_include,
--- a/voxblox/include/voxblox/io/layer_io_inl.h
+++ b/voxblox/include/voxblox/io/layer_io_inl.h
@@ -1,6 +1,8 @@
 #ifndef VOXBLOX_CORE_IO_LAYER_IO_INL_H_
 #define VOXBLOX_CORE_IO_LAYER_IO_INL_H_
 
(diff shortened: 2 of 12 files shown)


voxblox/include/voxblox/core/layer.h
```diff
@@ -225,9 +225,14 @@ class Layer {
   bool isCompatible(const LayerProto& layer_proto) const;
   bool isCompatible(const BlockProto& layer_proto) const;
   bool saveToFile(const std::string& file_path) const;
+  bool saveToFile(const std::string& file_path, bool clear_file) const;
```
ZacharyTaylor: do you need two functions, can't you just use a default value for clear_file?

voxblox/include/voxblox/io/layer_io_inl.h
```diff
@@ -0,0 +1,248 @@
+#ifndef VOXBLOX_CORE_IO_LAYER_IO_INL_H_
+#define VOXBLOX_CORE_IO_LAYER_IO_INL_H_
+
+#include <fstream>
+
+#include "./Block.pb.h"
```
ZacharyTaylor: I am really not a fan of relative file paths in includes, in this case couldn't this just be "Block.pb.h" anyway?

voxblox/src/utils/protobuf_utils.cc
```diff
@@ -31,7 +31,7 @@ bool writeProtoMsgCountToStream(uint32_t message_count,
   CHECK_NOTNULL(stream_out);
   CHECK(stream_out->is_open());
   stream_out->clear();
-  stream_out->seekg(0, std::ios::beg);
+  // stream_out->seekg(0, std::ios::beg);
```
ZacharyTaylor: why is this commented out, if it is unneeded shouldn't it just be removed

voxblox/include/voxblox/core/layer.h
helenol: Ah shit I remembered google style guide strongly discouraging default arguments, but checked it again and I'm wrong. Yeah here it makes waaaaaay more sense to do that. Changing.

voxblox/src/utils/protobuf_utils.cc
helenol: Good catch!

voxblox/include/voxblox/io/layer_io_inl.h
helenol: I think this specifies not to search anywhere else for this file (i.e., not the whole path) but rather look for it in this specific include path, since it's a locally-generated file? @mfehr I have no idea what the conventions around protobufs are. Help?

voxblox/test/test_protobuf.cc
```diff
@@ -211,9 +211,37 @@ TEST_F(ProtobufTsdfTest, LayerSubsetSerializationFromFile) {
...
+
+  voxblox::test::LayerTest<EsdfVoxel> esdf_test;
+
+  const std::string file = "multi_layer_test.voxblox";
+  bool clear_file = true;
+  io::SaveLayer(*layer_, file, clear_file);
+  clear_file = false;
+  io::SaveLayer(*esdf_layer, file, clear_file);
+
+  bool multiple_layer_support = true;
```
mfehr: constexpr

voxblox/test/test_protobuf.cc
```diff
@@ -211,9 +211,37 @@ TEST_F(ProtobufTsdfTest, LayerSubsetSerializationFromFile) {
...
+                                       &esdf_layer_from_file));
+
+  CompareLayers(*layer_, *tsdf_layer_from_file);
+  esdf_test.CompareLayers(*esdf_layer, *esdf_layer_from_file);
+}
+
 int main(int argc, char** argv) {
   ::testing::InitGoogleTest(&argc, argv);
   google::InitGoogleLogging(argv[0]);
+  FLAGS_alsologtostderr = 1;
```
mfehr: true

voxblox_ros/src/esdf_server.cc
```diff
@@ -139,6 +139,40 @@ void EsdfServer::publishPointclouds() {
...
+    this->esdf_map_pub_.publish(layer_msg);
+    publish_map_timer.Stop();
+  }
+
+  TsdfServer::publishMap();
+}
+
+bool EsdfServer::saveMap(const std::string& file_path) {
+  // Output TSDF map first, then ESDF.
+  bool success = TsdfServer::saveMap(file_path);
```
mfehr: const

voxblox_ros/src/esdf_server.cc
```diff
@@ -139,6 +139,40 @@ void EsdfServer::publishPointclouds() {
...
+
+  constexpr bool clear_file = false;
```
mfehr: constexpr variable naming

voxblox_ros/src/esdf_server.cc
```diff
@@ -139,6 +139,40 @@ void EsdfServer::publishPointclouds() {
...
+  return success &&
+         io::SaveLayer(esdf_map_->getEsdfLayer(), file_path, clear_file);
+}
+
+bool EsdfServer::loadMap(const std::string& file_path) {
+  // Load in the same order: TSDF first, then ESDF.
+  bool success = TsdfServer::loadMap(file_path);
```
mfehr: const

voxblox_ros/src/esdf_server.cc
```diff
@@ -139,6 +139,40 @@ void EsdfServer::publishPointclouds() {
...
+
+  constexpr bool multiple_layer_support = true;
```
mfehr: constexpr variable naming

voxblox_ros/src/tsdf_server.cc
```diff
@@ -389,6 +389,22 @@ bool TsdfServer::generateMesh() {
   return true;
 }
 
+bool TsdfServer::saveMap(const std::string& file_path) {
+  // Will only save TSDF layer for now. Inheriting classes should add other
```
mfehr: why would this save any other layer?

voxblox_ros/src/tsdf_server.cc
```diff
@@ -389,6 +389,22 @@ bool TsdfServer::generateMesh() {
...
+  // layers.
+  return io::SaveLayer(tsdf_map_->getTsdfLayer(), file_path);
+}
+
+bool TsdfServer::loadMap(const std::string& file_path) {
+  // Will only load TSDF layer for now. Inherited classes should add other
```
mfehr: same

voxblox_ros/src/tsdf_server.cc
```diff
@@ -389,6 +389,22 @@ bool TsdfServer::generateMesh() {
...
+  // layers.
+  constexpr bool multiple_layer_support = true;
```
mfehr: constexpr variable naming

voxblox/include/voxblox/io/layer_io_inl.h
```diff
@@ -0,0 +1,248 @@
...
+    } else if (!multiple_layer_support) {
+      LOG(ERROR)
+          << "The layer information read from file is not compatible with "
+             "the current layer!";
+      return false;
+    } else {
+      // Figure out how much offset to jump forward by. This is the number of
+      // blocks * the block size... Actually maybe we just read them in? This
+      // is probably easiest. Then if there's corrupted blocks, we abort.
+      // We just don't add these to the layer.
```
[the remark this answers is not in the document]
helenol: If you need that extra speed boost, why are you reading protobuf layers from file with multiple layer support? ;)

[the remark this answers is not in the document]
helenol: Also this way you avoid the problem you were really concerned about: being 1 off on the size of the skipped region, etc. I'd rather trade a bit of speed on a reading-from-disk function rather than lose robustness by doing crazy math.

voxblox/include/voxblox/io/layer_io_inl.h
```diff
@@ -0,0 +1,248 @@
...
+  } while (multiple_layer_support && !layer_found && !proto_file.eof());
+  return layer_found;
+}
+
+template <typename VoxelType>
+bool LoadBlocksFromFile(
+    const std::string& file_path,
+    typename Layer<VoxelType>::BlockMergingStrategy strategy,
+    Layer<VoxelType>* layer_ptr) {
+  bool multiple_layer_support = false;
```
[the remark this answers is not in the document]
helenol: It's checked in the function it calls below.

voxblox/include/voxblox/io/layer_io_inl.h
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    if ((layer_proto.voxel_size() <= 0.0f) ||
+        (layer_proto.voxels_per_side() == 0u)) {
+      LOG(ERROR)
+          << "Invalid parameter in layer protobuf message. Check the format.";
+      return false;
+    }
+
+    if (getVoxelType<VoxelType>().compare(layer_proto.type()) == 0) {
```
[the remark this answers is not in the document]
helenol: This is the function that allocates a new layer of the compatible layer type. The function that checks this checks also voxel size, vps, etc., etc., which we will just set correctly from the info in the file.

PR #150, opened 2018-02-02
late night attempt to fix transformation naming conventions with @mfehr

ffurrer:
We can probably also merge this back to master afterwards.

### DIFF
--- a/voxblox/include/voxblox/integrator/merge_integration.h
+++ b/voxblox/include/voxblox/integrator/merge_integration.h
@@ -90,16 +90,16 @@ void mergeLayerAintoLayerB(const Layer<VoxelType>& layer_A,
 // transformLayer for details
 template <typename VoxelType>
 void mergeLayerAintoLayerB(const Layer<VoxelType>& layer_A,
-                           const Transformation& T_A_B,
+                           const Transformation& T_B_A,
                            Layer<VoxelType>* layer_B,
                            bool use_naive_method = false) {
   Layer<VoxelType> layer_A_transformed(layer_B->voxel_size(),
                                        layer_B->voxels_per_side());
 
   if (use_naive_method) {
-    naiveTransformLayer(layer_A, T_A_B, &layer_A_transformed);
+    naiveTransformLayer(layer_A, T_B_A, &layer_A_transformed);
   } else {
-    transformLayer(layer_A, T_A_B, &layer_A_transformed);
+    transformLayer(layer_A, T_B_A, &layer_A_transformed);
   }
 
   mergeLayerAintoLayerB(layer_A_transformed, layer_B);
@@ -120,7 +120,7 @@ void resampleLayer(const Layer<VoxelType>& layer_in,
 // several orders of magnitude faster.
 template <typename VoxelType>
 void naiveTransformLayer(const Layer<VoxelType>& layer_in,
-                         const Transformation& T_in_out,
+                         const Transformation& T_out_in,
                          Layer<VoxelType>* layer_out) {
   BlockIndexList block_idx_list_in;
   layer_in.getAllAllocatedBlocks(&block_idx_list_in);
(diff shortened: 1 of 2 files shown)


voxblox/include/voxblox/integrator/merge_integration.h
```diff
@@ -139,7 +139,7 @@ void naiveTransformLayer(const Layer<VoxelType>& layer_in,
 
       // find voxel centers location in the output
       const Point voxel_center =
-          T_in_out *
```
mfehr: This was where we thought something was fishy, since the coordinates from `input_block.computeCoordinatesFromLinearIndex` are in the `in` frame, it doesn't make sense to  call `T_in_out * p_in`.

ZacharyTaylor: Are we taking transforms in voxblox to be active or passive?

ZacharyTaylor:
This PR actually brings up an issue I don't believe has been addressed. Do we have anywhere a clear lab wide doc giving variable naming and transform conventions. From looking round I have found:

* Some stuff in the MAV team wiki, though I wrote it when I first got here and on close examination it might be inconsistent
* The kalibr cheatsheet (though this doesn't go into variable naming). 
* There are also workshop slides from Paul Furgale in 2015 that give some good suggestions (https://static1.squarespace.com/static/523c5c56e4b0abc2df5e163e/t/53957839e4b05045ad65021d/1402304569659/Workshop+-+Rotations_v102.key.pdf). 
* And a brief page on the minkindr, aslam and maplab wiki https://github.com/ethz-asl/maplab/wiki/Expressing-frame-transformations-in-code

We also have the issue that most of our repos break at least some of the conventions outlined above (for example some use R for rotation matrices, others use C, if the frame of a point should go before or after the name, kalibr uses JPL quaternions, etc) 

If it doesn't already exist I would be in favor of having a meeting sometime soon to create a definitive ASL transformation convention wiki page that not only outlines what conventions to use but which repos deviate from it and how.

voxblox/include/voxblox/integrator/merge_integration.h
mfehr: To the best of my knowledge in voxblox we loosely follow the conventions of maplab/minkindr/aslam_cv, as described in your link below. The transformation uses minkindr and is therefore active. `T_in_out` should correspond to transforming point in the `out` frame to the `in` frame. There are some more examples here [1](https://github.com/ethz-asl/voxblox/blob/master/voxblox/src/integrator/tsdf_integrator.cc#L258), [2](https://github.com/ethz-asl/voxblox/blob/master/voxblox/src/integrator/tsdf_integrator.cc#L326),[3](https://github.com/ethz-asl/voxblox/blob/master/voxblox/src/integrator/tsdf_integrator.cc#L373)
But in terms of notation for points, vectors and such we haven't been following those very closely, e.g. we have points_C and points_W, and below I just added the frame to the variable to make it a bit clearer.

IMO in `transformLayer` and `naiveTransformLayer` the notation was wrong with respect to what the transformation was actually doing, the notation of everything on top of those function, was simply wrong because it was probably switched back and forth until it worked (same for the layer merging unit test i wrote) :)

mfehr:
@ZacharyTaylor I agree this PR hardly fixes all the notation inconsistencies, it only addresses this one error in the merge integration part and all it's consequences. (Although I think the rest of voxblox is pretty ok). Maybe you can bring this up in the next lab meeting?

voxblox/include/voxblox/integrator/merge_integration.h
ZacharyTaylor: I agree that the notation was wrong with respect to most of voxblox / maplab. I hadn't realized that minkindr was active as the docs for minkindr seem to favor passive representations (https://docs.leggedrobotics.com/kindr/page_rotations.html). 

LGTM

PR #155, opened 2018-02-28
Add compiling instructions.

marco-tranzatto:
without directly compiling "voxblox_rviz_plugin" the voxblox's mesh was not displayed in rviz.
Since "voxblox_rviz_plugin" is not a dependency of  "voxblox_ros" we need to compile it explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,12 @@ wstool set voxblox         --git https://github.com/ethz-asl/voxblox.git
 wstool update
 ```
 
+Compile:
+```
+cd ~/catkin_ws/src/
+catkin build voxblox_ros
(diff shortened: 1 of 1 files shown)


ZacharyTaylor:
I never noticed this, would it make sense to add a voxblox_rviz_plugin dependency to voxblox_ros?

marco-tranzatto:
Umh, maybe yes then. I could not visualize the mesh in rviz and I was wondering why ... It took me a couple of minutes to realize I didn't have the rviz plugin compiled.

marco-tranzatto:
If @helenol agrees I can open a new PR with the dependency and close this one.

helenol:
@marco-tranzatto helenol agrees with dependency being the better solution :)

PR #158, opened 2018-03-09
Merge bugfixes and features from GSM compatability branch

mfehr:
First of all: Sorry about the huge PR...
Second: big thanks to @ffurrer for his help with these fixes!

Changes:
 - More thorough unit testing for the indexing functions
 - Fix center voxel layer functions and their unit tests
 - Fix naming of the layer merging/transformation/interpolation logic
 - Move the mesh intergrator config out of the integrator to make it independent of the template
 - Make the mesh class more general and support normals
 - Fix bug in meshing
    - Corner case in zero crossing interpolation when the sdf values of two voxels are very close
    - Meshing at block boundaries. 
 - write vertex normals and alpha to PLY
 - fix some compilation warnings
 - Add additional functions to compare two layers, including an error layer output:
cow_optimized
scene
*Results of evaluating the alignment of two TSDF layers (cyan and magenta), TSDF errors of the overlapping voxels are visualized using colored points (red = high error, green = low error)* 

 - fix the TSDF visualization such that the user can specify the SDF range that is visualized (e.g. only close to surface within some threshold, and the colors are mapped correctly from - to + truncation distance) the rainbow colors only go from red to blue and green is actually at the zero crossing, not cyan.
image
*Visualization of the TSDF field, white dots represent the extracted surface, colored dots the TSDF values of the grid within a adjustable range around the surface (otherwise the pointclouds get to large)*

 - refactor and fix getConnectedMesh function that was suffering from integer overflows:
get_connected_mesh_after_bugfix
*Meshes resulting from getConnectedMesh before and after bugfix. Even though the bug was there in the implementation in master, there it only showed up at certain voxel sizes. The refactoring of the combineMesh() function made it more apparent, because the vertex similarity threshold was chosen much tighter.*

### DIFF
--- a/voxblox/include/voxblox/core/block.h
+++ b/voxblox/include/voxblox/core/block.h
@@ -53,6 +53,10 @@ class Block {
     return linear_index;
   }
 
+  // NOTE: This function is dangerous, it will truncate the voxel index to an
+  // index that is within this block if you pass a coordinate outside the range
+  // of this block. Try not to use this function if there is an alternative to
+  // directly address the voxels via precise integer indexing math.
   inline VoxelIndex computeVoxelIndexFromCoordinates(
       const Point& coords) const {
     const IndexElement max_value = voxels_per_side_ - 1;
@@ -65,6 +69,10 @@ class Block {
                       std::max(std::min(voxel_index.z(), max_value), 0));
   }
 
+  // NOTE: This function is dangerous, it will truncate the voxel index to an
+  // index that is within this block if you pass a coordinate outside the range
+  // of this block. Try not to use this function if there is an alternative to
+  // directly address the voxels via precise integer indexing math.
   inline size_t computeLinearIndexFromCoordinates(const Point& coords) const {
     return computeLinearIndexFromVoxelIndex(
         computeVoxelIndexFromCoordinates(coords));
@@ -103,6 +111,10 @@ class Block {
     return voxels_[computeLinearIndexFromVoxelIndex(index)];
   }
 
+  // NOTE: This function is dangerous, it will truncate the voxel index to an
+  // index that is within this block if you pass a coordinate outside the range
+  // of this block. Try not to use this function if there is an alternative to
+  // directly address the voxels via precise integer indexing math.
   inline const VoxelType& getVoxelByCoordinates(const Point& coords) const {
     return voxels_[computeLinearIndexFromCoordinates(coords)];
   }
@@ -157,6 +169,7 @@ class Block {
   FloatingPoint voxel_size_inv() const { return voxel_size_inv_; }
   size_t num_voxels() const { return num_voxels_; }
   Point origin() const { return origin_; }
+  void setOrigin(const Point& new_origin) { origin_ = new_origin; }
   FloatingPoint block_size() const { return block_size_; }
 
   bool has_data() const { return has_data_; }
@@ -193,7 +206,7 @@ class Block {
   // Base parameters.
   const size_t voxels_per_side_;
   const FloatingPoint voxel_size_;
-  const Point origin_;
+  Point origin_;
 
   // Derived, cached parameters.
   FloatingPoint voxel_size_inv_;
--- a/voxblox/include/voxblox/core/block_hash.h
+++ b/voxblox/include/voxblox/core/block_hash.h
@@ -1,8 +1,9 @@
-#ifndef VOXBLOX_BLOCK_HASH_H_
-#define VOXBLOX_BLOCK_HASH_H_
+#ifndef VOXBLOX_CORE_BLOCK_HASH_H_
+#define VOXBLOX_CORE_BLOCK_HASH_H_
 
 #include <functional>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 
 #include <Eigen/Core>
@@ -70,4 +71,4 @@ struct LongIndexHashMapType {
 
 }  // namespace voxblox
 
-#endif  // VOXBLOX_BLOCK_HASH_H_
+#endif  // VOXBLOX_CORE_BLOCK_HASH_H_
--- a/voxblox/include/voxblox/core/block_hash.h
+++ b/voxblox/include/voxblox/core/block_hash.h
@@ -29,8 +29,8 @@ struct AnyIndexHashMapType {
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
   typedef std::unordered_map<
-      BlockIndex, ValueType, AnyIndexHash, std::equal_to<BlockIndex>,
-      Eigen::aligned_allocator<std::pair<const BlockIndex, ValueType> > >
+      AnyIndex, ValueType, AnyIndexHash, std::equal_to<AnyIndex>,
+      Eigen::aligned_allocator<std::pair<const AnyIndex, ValueType> > >
       type;
 };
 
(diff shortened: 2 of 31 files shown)


voxblox/include/voxblox/core/block.h
```diff
@@ -157,6 +169,7 @@ class Block {
   FloatingPoint voxel_size_inv() const { return voxel_size_inv_; }
   size_t num_voxels() const { return num_voxels_; }
   Point origin() const { return origin_; }
+  void setOrigin(const Point& new_origin) { origin_ = new_origin; }
```
mfehr: needed if we want to transform/shift/move blocks.

voxblox/include/voxblox/core/layer.h
```diff
@@ -50,20 +50,18 @@ class Layer {
 
   inline const BlockType& getBlockByIndex(const BlockIndex& index) const {
     typename BlockHashMap::const_iterator it = block_map_.find(index);
-    if (it != block_map_.end()) {
-      return *(it->second);
-    } else {
+    if (it == block_map_.end()) {
```
mfehr: fixes warning

voxblox/include/voxblox/integrator/merge_integration.h
```diff
@@ -90,16 +91,16 @@ void mergeLayerAintoLayerB(const Layer<VoxelType>& layer_A,
 // transformLayer for details
 template <typename VoxelType>
 void mergeLayerAintoLayerB(const Layer<VoxelType>& layer_A,
-                           const Transformation& T_A_B,
+                           const Transformation& T_B_A,
```
mfehr: these changes in the transformation notation were reviewed and discussed in this PR: https://github.com/ethz-asl/voxblox/pull/150

voxblox/include/voxblox/mesh/marching_cubes.h
```diff
@@ -142,13 +142,16 @@ class MarchingCubes {
...
     const FloatingPoint sdf_diff = sdf1 - sdf2;
-    if (std::abs(sdf_diff) < min_diff) {
-      return Point(vertex1 + 0.5 * vertex2);
+    // Only compute the actual interpolation value if the sdf_difference is not
+    // too small, this is to counteract issues with floating point precision.
+    if (std::abs(sdf_diff) >= kMinSdfDifference) {
+      const FloatingPoint t = sdf1 / sdf_diff;
+      return Point(vertex1 + t * (vertex2 - vertex1));
+    } else {
+      return Point(0.5 * (vertex1 + vertex2));
```
mfehr: This is a bug, that was also there in openChisel. If the sdf difference is too small to properly divide by the value, in our opinion it should take point halfway in between vertex1 and vertex2, which translates to:
```
Point(0.5 * (vertex1 + vertex2))
```
Not
```
Point(vertex1 + 0.5 * vertex2
```
@ZacharyTaylor @helenol do you agree?

voxblox/include/voxblox/mesh/mesh.h
```diff
@@ -36,14 +37,25 @@ struct Mesh {
   typedef std::shared_ptr<Mesh> Ptr;
   typedef std::shared_ptr<const Mesh> ConstPtr;
 
+  static constexpr FloatingPoint kInvalidBlockSize = -1.0;
+
+  Mesh()
```
mfehr: If we combine all mesh-blocks into one global mesh, there is no point in using these block_size, origin and updated members.

voxblox/include/voxblox/mesh/mesh_integrator.h
```diff
@@ -42,19 +42,20 @@
 
 namespace voxblox {
 
+struct MeshIntegratorConfig {
```
mfehr: we moved this outside the mesh integrator, because otherwise this config struct needs to be templated on the VoxelType.

voxblox/include/voxblox/mesh/mesh_integrator.h
```diff
@@ -86,16 +87,19 @@ class MeshIntegrator {
...
+                 const Layer<VoxelType>& sdf_layer, MeshLayer* mesh_layer)
       : config_(config),
         sdf_layer_mutable_(nullptr),
         sdf_layer_const_(&sdf_layer),
         mesh_layer_(CHECK_NOTNULL(mesh_layer)) {
     initFromSdfLayer(sdf_layer);
 
-    cube_index_offsets_ << 0, 1, 1, 0, 0, 1, 1, 0, 0, 0, 1, 1, 0, 0, 1, 1, 0, 0,
-        0, 0, 1, 1, 1, 1;
+    // clang-format off
```
mfehr: only formatting

voxblox/include/voxblox/mesh/mesh_layer.h
```diff
@@ -136,18 +138,113 @@ class MeshLayer {
...
+
+    CHECK_EQ(combined_mesh->vertices.size(), combined_mesh->indices.size());
+  }
+
+  // Get a connected mesh by merging close vertices and removing triangles with
+  // zero surface area. If you only would like to connect vertices, make sure
+  // that the proximity threhsold <<< voxel size. If you would like to simplify
+  // the mesh, chose a threshold greater or near the voxel size until you
+  // reached the level of simpliciation desired.
+  void getConnectedMesh(
```
mfehr: TODO: There is still a bug in this function...

voxblox/test/test_tsdf_map.cc
```diff
@@ -31,48 +34,244 @@ TEST_F(TsdfMapTest, BlockAllocation) {
 }
 
 TEST_F(TsdfMapTest, IndexLookups) {
```
mfehr: Here we just added some more thorough unit testing for the indexing functions.

mfehr:
test this please

voxblox/src/io/mesh_ply.cc
```diff
@@ -51,38 +73,53 @@ bool outputMeshAsPly(const std::string& filename, const Mesh& mesh) {
...
+  if (mesh.hasNormals()) {
+    stream << "property float normal_x" << std::endl;
+    stream << "property float normal_y" << std::endl;
+    stream << "property float normal_z" << std::endl;
+  }
   if (mesh.hasColors()) {
     stream << "property uchar red" << std::endl;
     stream << "property uchar green" << std::endl;
     stream << "property uchar blue" << std::endl;
+    stream << "property uchar alpha" << std::endl;
```
helenol: crazy transparent meshes!!!

mfehr: doesn't really work in meshlab or cloud compare, but this way we at least don't lose the info :)

voxblox/include/voxblox/mesh/marching_cubes.h
ZacharyTaylor: Also agree

voxblox/src/io/mesh_ply.cc
ZacharyTaylor: Do you know if it works in rviz, I added it as a parameter when I wrote the rviz-plugin but never tested if it does anything

mfehr:
@ZacharyTaylor 
Regarding these random segfaults, I think I was confusing some things when I told you that. First of all, for gsm we don't use doubles (that's maplab) and second, I only noticed them when running all the unit tests and the segfault message was at random places. I think it was me who caused them in the test_tsdf_map unit test, but they were just showing up wherever in the log. Haven't seen them after fixing the test.

voxblox/include/voxblox/core/common.h
```diff
@@ -156,11 +160,15 @@ inline AnyIndex getGridIndexFromPoint(const Point& scaled_point) {
...
+// NOTE: This function is safer than `getGridIndexFromPoint`, because it assumes
+// we pass in not an arbitrary point in the grid cell, but the ORIGIN. This way
+// we can avoid the floating point precision issue that arrises for calls to
+// `getGridIndexFromPoint`for arbitrary points near the border of the grid cell.
 inline AnyIndex getGridIndexFromOriginPoint(const Point& point,
                                             const FloatingPoint grid_size_inv) {
-  return AnyIndex(std::round(point.x() * grid_size_inv),
-                  std::round(point.y() * grid_size_inv),
-                  std::round(point.z() * grid_size_inv));
+  return AnyIndex(std::floor(point.x() * grid_size_inv + 0.5),
```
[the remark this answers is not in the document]
mfehr: hmm, I'm positive that when I checked the diff, I had a good reason in mind why this was better, but I just tried to remember, but couldn't. So I'll undo the change :)

voxblox/include/voxblox/core/block_hash.h
```diff
@@ -44,6 +45,30 @@ typedef typename AnyIndexHashMapType<IndexSet>::type HierarchicalIndexSet;
...
+  }
+};
+
+template <typename ValueType>
+struct LongIndexHashMapType {
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+  typedef std::unordered_map<
+      LongIndex, ValueType, LongIndexHash, std::equal_to<LongIndex>,
+      Eigen::aligned_allocator<std::pair<const LongIndex, ValueType> > >
```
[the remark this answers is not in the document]
mfehr: Don't know, I didn't do any benchmarking, that's why I didn't just switch all hash maps, but we should definitely discuss this option.

voxblox/src/io/mesh_ply.cc
mfehr: Don't know, we didn't really use RVIZ to visualize results, margarita has her own PCL based viewer for GSM and in modelify we dumped data to ply to look at it.

mfehr:
Thanks for the reviews @ffurrer @helenol @ZacharyTaylor !

PR #161, opened 2018-03-15
Mark hallucinated voxels from addNewRobotPosition.

helenol:
This fixes the issue that we can't start voxblox mapping while the drone is on the ground, and then its current position is marked as occupied in the map.
Now, all voxels added to the ESDF within the bounding spheres for marking unknown space as either occupied or free are marked as "hallucinated". This allows them to be cleared by future calls to addNewRobotPosition, making sure that the current pose of the drone, even if never observed, is always considered free. (If observed, of course, this takes priority).

As a note, the hallucination status isn't serialized, (1) because I don't think it's necessary to serialize and (2) because it's hard to do that without breaking back-compatibility. Then again if no one other than me serializes ESDFs....

### DIFF
--- a/voxblox/include/voxblox/core/voxel.h
+++ b/voxblox/include/voxblox/core/voxel.h
@@ -24,6 +24,9 @@ struct EsdfVoxel {
   bool observed = false;
   bool in_queue = false;
   bool fixed = false;
+  // Whether the voxel was copied from the TSDF (false) or created from a pose
+  // or some other source (true).
+  bool hallucinated = false;
   // Relative direction toward parent. If itself, then either uninitialized
   // or in the fixed frontier.
   Eigen::Vector3i parent = Eigen::Vector3i::Zero();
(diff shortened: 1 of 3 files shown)


voxblox/include/voxblox/core/voxel.h
```diff
@@ -24,6 +24,9 @@ struct EsdfVoxel {
   bool observed = false;
   bool in_queue = false;
   bool fixed = false;
+  // Whether the voxel was copied from the TSDF (false) or created from a pose
+  // or some other source (true).
+  bool hallucinated = false;
```
mfehr: add a comment to non-serializable members?

voxblox/src/integrator/esdf_integrator.cc
```diff
@@ -65,9 +65,11 @@ void EsdfIntegrator::addNewRobotPosition(const Point& position) {
         continue;
       }
       EsdfVoxel& esdf_voxel = block_ptr->getVoxelByVoxelIndex(voxel_index);
-      if (!esdf_voxel.observed) {
+      // We can clear unobserved or hallucinated voxels.
+      if (!esdf_voxel.observed || esdf_voxel.hallucinated) {
         esdf_voxel.distance = config_.default_distance_m;
         esdf_voxel.observed = true;
+        esdf_voxel.hallucinated = true;
```
mfehr: maybe consider adding this new behavior to the description/comment of this function?

PR #163, opened 2018-03-21
version/double precision

mfehr:
Do not merge into master, this PR tracks the latest double precision version of voxblox used in maplab_private.


mfehr:
Will no longer be necessary for maplab, we switched to floats as well.

PR #164, opened 2018-03-21
Merge changes from double precision branch into master

mfehr:
There are some minor changes in the double precision branch. We don't need it anymore and are using the voxblox master version in maplab, therefore it's time to clean this up.


voxblox/include/voxblox/io/sdf_ply.h
```diff
@@ -31,18 +31,18 @@ enum PlyOutputTypes {
 // SDF values within this max value result in a colored point.
 template <typename VoxelType>
 bool getColorFromVoxel(const VoxelType& voxel,
-                       const FloatingPoint sdf_color_range,
-                       const FloatingPoint sdf_max_value, Color* color);
+                       const float sdf_color_range,
```
helenol: Ahhh right, since the SDF distances are always floats anyway. This makes more sense.

voxblox/include/voxblox/io/sdf_ply.h
```diff
@@ -142,19 +142,19 @@ bool convertLayerToMesh(
...
 // Output the layer to ply file. Depending on the ply output type, this either
 // exports all voxel centers colored by th SDF values or extracts the ISO
-// surface as mesh. The parameter max_sdf_range is used to color the points for
+// surface as mesh. The parameter sdf_color_range is used to color the points for
 // modes that use an SDF-based point cloud coloring function.
 template <typename VoxelType>
 bool outputLayerAsPly(const Layer<VoxelType>& layer,
                       const std::string& filename, PlyOutputTypes type,
-                      const FloatingPoint max_sdf_range = 0.3,
+                      const FloatingPoint sdf_color_range = 0.3,
```
helenol: Should this also be straight floats here as well?

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -335,9 +335,9 @@ void MergedTsdfIntegrator::bundleRays(
     }
   }
 
-  LOG(INFO) << "Went from " << points_C.size() << " points to "
-            << voxel_map->size() << " raycasts  and " << clear_map->size()
-            << " clear rays.";
+  VLOG(3) << "Went from " << points_C.size() << " points to "
```
helenol: Thanks <3

voxblox/include/voxblox/io/sdf_ply.h
mfehr: yes you're right, I guess I just stopped changing them when the compiler was happy :P

PR #167, opened 2018-04-13
adding checks to ensure preconditions

mbuerki:
Some mistake in the calling code to the tsdf integrator (with num points != num colors) resulted in a segfault in voxblox. I therefore added the checks below, because this condition (num points == num colors) is a hard assumption in this part of the code, and leads to the said segfault if violated.


helenol:
add to whitelist

PR #168, opened 2018-04-13
Remove voxblox node, use tsdf_server and esdf_server instead.

helenol:
Documentation is also updated.
For 90% of use-cases, tsdf_server is a drop-in replacement for voxblox node. Parameters should be the same.


salihmarangoz:
**output_mesh_as_pcl_mesh** and **output_mesh_as_pointcloud** are completely gone with this commit ;_;

PR #169, opened 2018-04-15
Tsdf/Esdf Server - Clear map service call and additional serialization options

timethy:
This adds a clear map service call.

**Help wanted**
There seems to be a bug however:
When I run this with mav_local_avoidance (firefly_sim_all.launch) then call the clear_map service of the voxblox_node (`rosservice call /firefly/voxblox_node/clear_map`, 
it crashes the lcto_planner node, presumably because the new esdf it receives has less blocks than it expect it to be.
I didn't dug deep enough into the code yet to fully understand what's happening.
The error output is:
```
conversions_inl.h:69] Check failed: layer->getNumberOfAllocatedBlocks() == msg.blocks.size() (61 vs. 47) 
*** Check failure stack trace: ***
    @     0x7fbc6a55226d  google::LogMessage::Fail()
    @     0x7fbc6a554083  google::LogMessage::SendToLog()
    @     0x7fbc6a551dfb  google::LogMessage::Flush()
    @     0x7fbc6a554a6e  google::LogMessageFatal::~LogMessageFatal()
    @     0x7fbc6b875577  voxblox::deserializeMsgToLayer<>()
    @     0x7fbc6b867776  voxblox::EsdfServer::esdfMapCallback()
    @     0x7fbc6b8759fc  ros::SubscriptionCallbackHelperT<>::call()
    @     0x7fbc6b191e2d  ros::SubscriptionQueue::call()
    @     0x7fbc6b1376f8  ros::CallbackQueue::callOneCB()
    @     0x7fbc6b1390fb  ros::CallbackQueue::callAvailable()
    @     0x7fbc6b195ef9  ros::SingleThreadedSpinner::spin()
    @     0x7fbc6b17aedb  ros::spin()
    @           0x422925  main
    @     0x7fbc692c6830  __libc_start_main
    @           0x422a49  _start
```

which means that this check is failing (line 69 in voxblox_ros/conversions_inl.h)
```
CHECK_EQ(layer->getNumberOfAllocatedBlocks(), msg.blocks.size());
```

@helenol maybe you have an idea what's going wrong?


helenol:
add to whitelist

mfehr:
@timethy I'm on it, will push some changes in a min.

mfehr:
@helenol ptal
@timethy please give this a try and check if the planner is still functioning after this.

voxblox_msgs/msg/Layer.msg
```diff
@@ -9,5 +9,11 @@ uint8 action   # See action defines below
 voxblox_msgs/Block[] blocks
 
 # Action definitions
-uint8 ACTION_FULL_MAP = 0
+# Set the layer to the state described by this message.
+uint8 ACTION_RESET = 0
```
helenol: Hmm... Are you use reset should be the default action? I'd prefer merge as 0 and reset as 2, for instance. that way if it's not set, it does something reasonable.

timethy:
@mfehr Just tested this and it works great!

voxblox_msgs/msg/Layer.msg
mfehr: The default action that is set during serialization is KUpdate, but you're right, if someone just initializes the ROS msg the default is reset. On the other hand this is the way it was before, FULL MAP = RESET, I thought you might have rosbags somewhere that need to remain compatible?

helenol: Not sure there's many rosbags where we actually record the map, and I wouldn't worry about invalidating them. If that's the only concern, tehn I'd prefer to swap` 0 <-> 2`

mfehr: Ok, if you're not concerned about this then I'll change it :)

PR #176, opened 2018-06-12
Add PLY output to TSDF viewer, change to standard ROS messaging.

helenol:
Should be more of a standard ROS node now. ;)


ZacharyTaylor:
LGTM

This change to me actually raises the question of what we gain by depending on glog.
Even in the voxblox core library that should remain ros independent we could use [console_bridge](http://wiki.ros.org/console_bridge) and do the [same magic cuckoo_time_translator does](https://github.com/ethz-asl/cuckoo_time_translator/blob/master/cuckoo_time_translator/src/RosConsoleBridgeActivator.cpp) to allow it to output on /rosout when it is part of voxblox_ros.

This would ensure all logging is done in a consistent manner throughout the whole lib

PR #186, opened 2018-07-25
Skeleton + Thermal + ARCHE changes

helenol:
This is an unfortunately massive PR, most of which doesn't affect the normal use-cases at all.

* Changes to core stuff
  * Change raycasting and global voxel indices to long indices, so we don't run into problems with this later.
  * Add a const version of getVoxelPtrByCoordinates in the block
  * Add getVoxelPtrByCoordinates to the layer (needed this for a long time...)
  * Fix gray colormap
  * Add more colormaps and allow lookup in them by setting min and max in the colormap object
* New voxel types
  * Intensity voxel - stores temperature
    *  Intensity integrator - uses existing geometry to project temperature on it
* Utilities
  * Add neighbor look-up tools
  * Add distance look-up tools

The skeleton stuff will live in mav_voxblox_planning instead.


voxblox/include/voxblox/skeletons/neighbor_tools.h
```diff
@@ -0,0 +1,177 @@
...
+
+template <typename VoxelType>
+Eigen::Vector3i NeighborTools<VoxelType>::getOffsetBetweenVoxels(
+    const BlockIndex& start_block_index, const VoxelIndex& start_voxel_index,
+    const BlockIndex& end_block_index,
+    const VoxelIndex& end_voxel_index) const {
+  BlockIndex current_block_index = end_block_index;
+  Eigen::Vector3i voxel_offset = Eigen::Vector3i::Zero();
+
+  // Line up the blocks so that they're in the same block.
```
ZacharyTaylor: this comment doesn't make sense

voxblox/include/voxblox/skeletons/neighbor_tools.h
```diff
@@ -0,0 +1,177 @@
...
+  for (unsigned int i = 0; i < 3; ++i) {
+    while (start_block_index(i) > current_block_index(i)) {
```
ZacharyTaylor: Am I missing something why does this need to be a loop, can't we just get everything from turning the local indexes into global indexes and then taking the difference?

voxblox/include/voxblox/skeletons/skeleton.h
```diff
@@ -0,0 +1,135 @@
...
+  // Access to all the skeleton points.
+  const AlignedVector<SkeletonPoint>& getSkeletonPoints() const {
+    return points_;
+  }
+  const AlignedVector<SkeletonPoint>& getEdgePoints() const { return edges_; }
+  const AlignedVector<SkeletonPoint>& getVertexPoints() const {
+    return vertices_;
+  }
+
+  AlignedVector<SkeletonPoint>& getSkeletonPoints() { return points_; }
```
ZacharyTaylor: If you are offering full access to all the private variables, why not just make this a struct with these members public?

voxblox/include/voxblox/skeletons/skeleton.h
```diff
@@ -0,0 +1,135 @@
...
+ private:
+  AlignedVector<SkeletonPoint> points_;
+
+  // Subsets of points (just copies because we don't care)
+  AlignedVector<SkeletonPoint> vertices_;
+  AlignedVector<SkeletonPoint> edges_;
+};
+
+struct SkeletonVertex {
+  int64_t vertex_id = -1;
```
ZacharyTaylor: Are these negitive defaults needed? Is there a case where you will create a Vetrex but not assign the data?

voxblox/include/voxblox/core/voxel.h
```diff
@@ -32,21 +28,37 @@ struct EsdfVoxel {
...
+struct SkeletonVoxel {
+  float distance = 0.0f;
+  uint8_t num_basis_points = 0u;
+  bool is_face = false;
+  bool is_edge = false;
+  bool is_vertex = false;
+  int64_t vertex_id = -1;
+};
+
+struct ThermalVoxel {
```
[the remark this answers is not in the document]
mfehr: I agree, could be useful for laser too. The observations field also seems tailored to your particular application, it can only store a very low number of observations, which I assume corresponds to some threshold you use (> 0?). So IMO either replace it with a bool, turn it into a weight or drop it (in which case you can set the intensity value to an invalid value to represent unobserved voxels)?
@helenol  @ZacharyTaylor wdyt?

voxblox/include/voxblox/integrator/thermal_integrator.h
```diff
@@ -0,0 +1,51 @@
...
+
+class ThermalIntegrator {
+ public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+  ThermalIntegrator(const Layer<TsdfVoxel>& tsdf_layer,
+                    Layer<ThermalVoxel>* thermal_layer);
+
+  // Set the max distance for projecting into the TSDF layer.
+  void setMaxDistance(FloatingPoint max_distance) {
```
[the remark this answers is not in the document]
helenol: Default is set in the constructor, this is just an optional overwrite. :)

voxblox/include/voxblox/skeletons/nanoflann_interface.h
```diff
@@ -0,0 +1,73 @@
+#ifndef VOXBLOX_SKELETONS_NANOFLANN_INTERFACE_H_
```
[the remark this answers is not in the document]
helenol: While I agree with you on it being much cleaner, I'm concerned about cross-linking dependencies with other versions of nanoflann. I'm not sure opencv or PCL push it into its own namespace/headerguard, and I'm not sure how to do that just from the CMake lists -- then there's the potential of interfering with the versions brought in from those libraries if cross-linked. This was the original reason for just adding a new version in. @ZacharyTaylor, wdyt?

voxblox/include/voxblox/skeletons/neighbor_tools.h
```diff
@@ -0,0 +1,177 @@
...
+
+// Uses 26-connectivity and quasi-Euclidean distances.
+// Directions is the direction that the neighbor voxel lives in. If you
+// need the direction FROM the neighbor voxel TO the current voxel, take
+// negative of the given direction.
+template <typename VoxelType>
+void NeighborTools<VoxelType>::getNeighborsAndDistances(
+    const BlockIndex& block_index, const VoxelIndex& voxel_index,
+    int connectivity, AlignedVector<VoxelKey>* neighbors,
+    AlignedVector<float>* distances,
```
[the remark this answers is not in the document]
helenol: Distances have always been floats inside the voxels, regardless of FloatingPoint definitions. Only positions, not voxel-stored distances, were different.

voxblox/include/voxblox/core/voxel.h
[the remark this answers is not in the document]
helenol: Sounds good, I'll switch to IntensityVoxel and add serialization.

voxblox/include/voxblox/skeletons/neighbor_tools.h
```diff
@@ -0,0 +1,177 @@
...
+
+namespace voxblox {
+
+template <typename VoxelType>
+class NeighborTools {
+ public:
+  NeighborTools() : voxels_per_side_(0) {}
+  NeighborTools(const Layer<VoxelType>* layer) : layer_(layer) {
+    CHECK_NOTNULL(layer);
+    voxels_per_side_ = layer_->voxels_per_side();
```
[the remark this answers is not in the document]
helenol: Just cached for faster/shorter access since it's used often.

voxblox/include/voxblox/skeletons/neighbor_tools.h
```diff
@@ -0,0 +1,177 @@
...
+      int connectivity, AlignedVector<VoxelKey>* neighbors,
+      AlignedVector<float>* distances,
+      AlignedVector<Eigen::Vector3i>* directions) const;
+
+  void getNeighbor(const BlockIndex& block_index, const VoxelIndex& voxel_index,
+                   const Eigen::Vector3i& direction,
+                   BlockIndex* neighbor_block_index,
+                   VoxelIndex* neighbor_voxel_index) const;
+
+  Eigen::Vector3i getOffsetBetweenVoxels(
```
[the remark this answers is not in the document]
helenol: Good call, added and replaced.

voxblox/include/voxblox/skeletons/neighbor_tools.h
```diff
@@ -0,0 +1,177 @@
...
+    const BlockIndex& block_index, const VoxelIndex& voxel_index,
+    int connectivity, AlignedVector<VoxelKey>* neighbors,
+    AlignedVector<float>* distances,
+    AlignedVector<Eigen::Vector3i>* directions) const {
+  CHECK_NOTNULL(layer_);
+  CHECK_NOTNULL(neighbors);
+  CHECK_NOTNULL(distances);
+  CHECK_NOTNULL(directions);
+
+  static const double kSqrt2 = std::sqrt(2);
```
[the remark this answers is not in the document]
helenol: Can't do with std::sqrt :( https://stackoverflow.com/questions/8622256/in-c11-is-sqrt-defined-as-constexpr

voxblox/src/integrator/intensity_integrator.cc
```diff
@@ -0,0 +1,67 @@
...
+    // then fill in their value. Otherwise don't.
+    Point close_voxel = surface_intersection;
+    for (int voxel_offset = -intensity_prop_voxel_radius;
+         voxel_offset <= intensity_prop_voxel_radius; voxel_offset++) {
+      close_voxel =
+          surface_intersection + bearing_vectors[i] * voxel_offset * voxel_size;
+      Block<IntensityVoxel>::Ptr new_block_ptr =
+          intensity_layer_->allocateBlockPtrByCoordinates(close_voxel);
+      IntensityVoxel& new_voxel = block_ptr->getVoxelByCoordinates(close_voxel);
+      if (new_voxel.weight == 0) {
```
mfehr: the weight is now a float and shouldn't be compared to 0 this way.

voxblox/test/test_approx_hash_array.cc
```diff
@@ -109,6 +109,57 @@ TEST_F(ApproxHashArrayTest, SetRandomWriteRead) {
...
 }
 
+TEST_F(ApproxHashArrayTest, SetRandomWriteReadLongIndex) {
+  // storing 65536 size_ts
+  ApproxHashSet<16, 10, LongIndex, LongIndexHash> approx_hash_set;
+
+  // generate 1000 random elements
+  AlignedVector<LongIndex> rand_indexes;
+  std::mt19937 gen(1);
+  std::uniform_int_distribution<int64_t> dis(1, 10000000000000000);
```
mfehr: would it make sense to use the max range from [std::numeric_limits](https://en.cppreference.com/w/cpp/types/numeric_limits) ? (same in the old unit test you adapted?)

voxblox/test/test_approx_hash_array.cc
```diff
@@ -109,6 +109,57 @@ TEST_F(ApproxHashArrayTest, SetRandomWriteRead) {
...
+  for (int i = 0; i < 1000; ++i) {
+    rand_indexes.push_back(LongIndex(dis(gen), dis(gen), dis(gen)));
+  }
+
+  // test insert and clearing
+  for (size_t i = 0; i < 50u; ++i) {
+    approx_hash_set.resetApproxSet();
+
+    // insert their values
+    int inserted = 0;
```
mfehr: unsigned type? and 950u below

voxblox/test/test_protobuf.cc
```diff
@@ -133,6 +134,38 @@ TEST_F(ProtobufEsdfTest, LayerSerialization) {
   CompareLayers(*layer_, layer_from_proto);
 }
 
+TEST_F(ProtobufIntensityTest, BlockSerialization) {
```
mfehr: +1 !

voxblox_ros/include/voxblox_ros/intensity_server.h
```diff
@@ -0,0 +1,55 @@
+#ifndef VOXBLOX_ROS_intensity_SERVER_H_
+#define VOXBLOX_ROS_intensity_SERVER_H_
+
+#include <memory>
+#include <cv_bridge/cv_bridge.h>
```
mfehr: nit: split include groups (or alph)

voxblox_ros/include/voxblox_ros/intensity_vis.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <voxblox/utils/color_maps.h>
+#include <voxblox_msgs/Mesh.h>
+
+#include "voxblox_ros/conversions.h"
+#include "voxblox_ros/mesh_vis.h"
+
+namespace voxblox {
+
+inline void recolorVoxbloxMeshMsgByIntensity(
```
mfehr: Is there a reason why is the color not applied at a lower level, i.e. to the voxblox::Mesh ?

voxblox_ros/include/voxblox_ros/intensity_vis.h
```diff
@@ -0,0 +1,42 @@
+#ifndef VOXBLOX_ROS_THERMAL_VIS_H_
```
mfehr: Fix header guards. Did you disable the linter? There are quite a few things in this PR that shouldn't have gotten through the linter :| Like the signed/unsigned comparisons, lots of missing system includes, ref to temporary, header guards, etc.

voxblox_ros/include/voxblox_ros/tsdf_server.h
```diff
@@ -127,8 +127,12 @@ class TsdfServer {
   // What output information to publish
   bool publish_tsdf_info_;
   bool publish_slices_;
+  bool publish_pointclouds_;
   bool publish_tsdf_map_;
 
+  // Whether to save the latest mesh message sent (for inheriting classes).
+  bool cache_mesh_;
```
mfehr: nit: `cached_mesh_`?

voxblox_ros/include/voxblox_ros/tsdf_server.h
```diff
@@ -167,6 +171,8 @@ class TsdfServer {
   // Mesh accessories.
   std::shared_ptr<MeshLayer> mesh_layer_;
   std::unique_ptr<MeshIntegrator<TsdfVoxel>> mesh_integrator_;
+  // Optionally cached mesh message.
+  voxblox_msgs::Mesh mesh_msg_;
```
mfehr: nit: `cached_mesh_msg_`?

Group the two mesh caches?

voxblox_ros/src/esdf_server.cc
```diff
@@ -120,7 +117,11 @@ void EsdfServer::updateMesh() {
     const bool clear_updated_flag_esdf = false;
     esdf_integrator_->updateFromTsdfLayer(clear_updated_flag_esdf);
   }
-  publishAllUpdatedEsdfVoxels();
+  if (publish_pointclouds_) {
+    publishAllUpdatedEsdfVoxels();
+  } else if (publish_traversable_) {
```
mfehr: why is this an either or? The flags don't really tell and not sure this is intuitive.

voxblox_ros/src/intensity_server.cc
```diff
@@ -0,0 +1,123 @@
+#include "voxblox_ros/intensity_server.h"
+
+namespace voxblox {
+
+IntensityServer::IntensityServer(const ros::NodeHandle& nh,
+                                 const ros::NodeHandle& nh_private)
+    : TsdfServer(nh, nh_private),
+      focal_length_px_(391.5f),
```
mfehr: seems rather specific? Not sure it makes sense that there is a default value for the focal length? I think this should be a mandatory ros param and crash and burn if not provided.

voxblox_ros/src/intensity_server.cc
```diff
@@ -0,0 +1,123 @@
...
+                                tsdf_map_->getTsdfLayer().voxels_per_side()));
+  intensity_integrator_.reset(new IntensityIntegrator(tsdf_map_->getTsdfLayer(),
+                                                      intensity_layer_.get()));
+
+  // Get ROS params:
+  nh_private_.param("intensity_focal_length", focal_length_px_,
+                    focal_length_px_);
+  nh_private_.param("subsample_factor", subsample_factor_, subsample_factor_);
+
+  float intensity_min_value = 10.0;
```
mfehr: nit: 10.f and below

voxblox_ros/src/intensity_server.cc
```diff
@@ -0,0 +1,123 @@
...
+  intensity_image_sub_ = nh_private_.subscribe(
+      "intensity_image", 1, &IntensityServer::intensityImageCallback, this);
+}
+
+void IntensityServer::updateMesh() {
+  TsdfServer::updateMesh();
+
+  // Now recolor the mesh...
+  timing::Timer publish_mesh_timer("intensity_mesh/publish");
+  recolorVoxbloxMeshMsgByIntensity(*intensity_layer_, color_map_, &mesh_msg_);
```
mfehr: Oh I see, so you just overwrite the cached mesh of the TsdfServer. I guess that's probably more efficient than repainting and then reconverting the underlying voxblox mesh type.

voxblox_ros/src/intensity_server.cc
```diff
@@ -0,0 +1,123 @@
...
+  pointcloud.header.frame_id = world_frame_;
+  intensity_pointcloud_pub_.publish(pointcloud);
+
+  TsdfServer::publishPointclouds();
+}
+
+void IntensityServer::intensityImageCallback(
+    const sensor_msgs::ImageConstPtr& image) {
+  CHECK(intensity_layer_);
+  CHECK(intensity_integrator_);
```
mfehr: add CNN for image

voxblox_ros/src/intensity_server.cc
```diff
@@ -0,0 +1,123 @@
...
+                                    image->header.stamp, &T_G_C)) {
+    ROS_WARN_THROTTLE(10, "Failed to look up intensity transform!");
+    return;
+  }
+
+  cv_bridge::CvImageConstPtr cv_ptr = cv_bridge::toCvShare(image);
+
+  CHECK(cv_ptr);
+
+  size_t num_pixels =
```
mfehr: const here and below

voxblox_ros/src/intensity_server.cc
```diff
@@ -0,0 +1,123 @@
...
+      cv_ptr->image.rows * cv_ptr->image.cols / subsample_factor_;
+
+  float half_row = cv_ptr->image.rows / 2.0;
+  float half_col = cv_ptr->image.cols / 2.0;
+
+  // Pre-allocate the bearing vectors and temperatures.
+  Pointcloud bearing_vectors;
+  bearing_vectors.reserve(num_pixels + 1);
```
mfehr: why the +1?

voxblox_ros/src/intensity_server.cc
```diff
@@ -0,0 +1,123 @@
...
+
+  size_t k = 0;
+  size_t m = 0;
+  for (int i = 0; i < cv_ptr->image.rows; i++) {
+    const float* image_row = cv_ptr->image.ptr<float>(i);
+    for (int j = 0; j < cv_ptr->image.cols; j++) {
+      if (m % subsample_factor_ == 0) {
+        bearing_vectors.push_back(
+            T_G_C.getRotation().toImplementation() *
+            Point(j - half_col, i - half_row, focal_length_px_).normalized());
```
mfehr: I guess here it would make sense to add a local loop variable for better readability (it's gonna be optimized away anyway.)

voxblox/include/voxblox/utils/distance_utils.h
```diff
@@ -0,0 +1,79 @@
...
+        layer.getBlockPtrByCoordinates(current_pos);
+    if (!block_ptr) {
+      // How much should we move up by? 1 voxel? 1 block? Could be close to the
+      // block boundary though....
+      // Let's start with the naive choice: 1 voxel.
+      t += voxel_size;
+      continue;
+    }
+    const VoxelType& voxel = block_ptr->getVoxelByCoordinates(current_pos);
+    if (voxel.weight < 1e-6) {
```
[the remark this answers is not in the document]
helenol: Yes, which is desired behavior. This is for projecting onto the closest surface along a bearing vector, regardless of the state of space inbetween.

voxblox/include/voxblox/utils/neighbor_tools.h
```diff
@@ -0,0 +1,181 @@
...
+// Directions is the direction that the neighbor voxel lives in. If you
+// need the direction FROM the neighbor voxel TO the current voxel, take
+// negative of the given direction.
+template <typename VoxelType>
+void NeighborTools<VoxelType>::getNeighborsAndDistances(
+    const BlockIndex& block_index, const VoxelIndex& voxel_index,
+    Connectivity connectivity, AlignedVector<VoxelKey>* neighbors,
+    AlignedVector<float>* distances,
+    AlignedVector<SignedIndex>* directions) const {
+  CHECK_NOTNULL(layer_);
```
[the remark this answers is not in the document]
helenol: Yes it is, you need to know the voxel size, vps, etc., etc. Much easier to just get it directly from the layer object so you're sure it's set up correctly.

voxblox/include/voxblox/utils/neighbor_tools.h
```diff
@@ -0,0 +1,181 @@
...
+#include "voxblox/core/common.h"
+#include "voxblox/core/layer.h"
+
+namespace voxblox {
+
+// Define what connectivity you want to have in the voxels.
+enum Connectivity { kSix = 0, kEighteen, kTwentySix };
+
+template <typename VoxelType>
+class NeighborTools {
```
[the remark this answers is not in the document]
helenol: I think it's a huge pain in the ass to do neighbor computations often if you have to get the layer, get the voxel size, get the vps, etc. every single time. I'd rather have a utils class you can use everywhere and just cache the stuff.

voxblox/include/voxblox/utils/neighbor_tools.h
```diff
@@ -0,0 +1,181 @@
...
+    for (unsigned int i = 0; i < 3; ++i) {
+      unsigned int next_i = (i + 1) % 3;
+      for (int j = -1; j <= 1; j += 2) {
+        direction(i) = j;
+        for (int k = -1; k <= 1; k += 2) {
+          direction(next_i) = k;
+          getNeighbor(block_index, voxel_index, direction, &neighbor.first,
+                      &neighbor.second);
+          neighbors->emplace_back(neighbor);
+          distances->emplace_back(kSqrt2);
```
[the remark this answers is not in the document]
helenol: 1. Zec's last review asked me to change the int connectivity value to an enum wiht 3 values since there are only 3 valid values. I agree there, I think it makes more sense to impose only proper connectivity option so that the user isn't surprised to get a number different from what they put in.
2. I think this is a lot worse for readability.

voxblox/include/voxblox/utils/neighbor_tools.h
```diff
@@ -0,0 +1,181 @@
...
+    AlignedVector<SignedIndex>* directions) const {
+  CHECK_NOTNULL(layer_);
+  CHECK_NOTNULL(neighbors);
+  CHECK_NOTNULL(distances);
+  CHECK_NOTNULL(directions);
+
+  static const float kSqrt2 = std::sqrt(2);
+  static const float kSqrt3 = std::sqrt(3);
+
+  neighbors->reserve(connectivity);
```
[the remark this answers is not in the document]
helenol: Ah whooooops good catch. Yeah this is from before I switched the enums when connectivity was just an int.

voxblox/src/integrator/intensity_integrator.cc
```diff
@@ -0,0 +1,67 @@
...
+      intensity_layer_(intensity_layer) {}
+
+void IntensityIntegrator::addIntensityBearingVectors(
+    const Point& origin, const Pointcloud& bearing_vectors,
+    const std::vector<float>& intensities) {
+  timing::Timer intensity_timer("intensity/integrate");
+
+  CHECK_EQ(bearing_vectors.size(), intensities.size())
+      << "Intensity and bearing vector size does not match!";
+  const int intensity_prop_voxel_radius = 2;
```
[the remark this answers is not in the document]
helenol: NO GFLAGS IN VOXBLOX!!!!!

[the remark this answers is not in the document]
mfehr: Haha I forgot, sorry for bringing up this sensitive topic :D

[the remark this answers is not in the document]
helenol: But switching to parameter. ;)

voxblox/src/integrator/intensity_integrator.cc
```diff
@@ -0,0 +1,67 @@
...
+  const FloatingPoint voxel_size = tsdf_layer_.voxel_size();
+
+  for (size_t i = 0; i < bearing_vectors.size(); ++i) {
+    Point surface_intersection = Point::Zero();
+    // Cast ray from the origin in the direction of the bearing vector until
+    // finding an intersection with a surface.
+    bool success = getSurfaceDistanceAlongRay<TsdfVoxel>(
```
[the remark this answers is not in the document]
mfehr: Oh right, I actually didn't realize you are stepping ahead by the distance in the voxel, smart! Yeah in that case obviously this is faster. Ignore my comment then.

voxblox/include/voxblox/utils/neighbor_tools.h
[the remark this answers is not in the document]
mfehr: if that's the case then yes, but it isn't, I checked. The only variable you even extract and use is vps and that hardly justifies having a class and storing a pointer to the layer that you never use.

voxblox/include/voxblox/utils/neighbor_tools.h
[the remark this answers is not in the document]
mfehr: see comment above :)

voxblox/include/voxblox/utils/distance_utils.h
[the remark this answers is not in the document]
mfehr: I see, nvm then.

voxblox/test/test_protobuf.cc
helenol: I'm very slowly learning...

voxblox_ros/include/voxblox_ros/intensity_vis.h
helenol: Easier to do it, I guess? Fewer problems with consistency when updating the mesh incrementally and you don't have to keep 2 full copies around?

voxblox_ros/include/voxblox_ros/tsdf_server.h
helenol: Nope, it's a present tense verb as it's a setting: whether to *cache* the mesh, not whether it's already been cached.

voxblox_ros/src/esdf_server.cc
helenol: Good catch!

voxblox_ros/src/intensity_server.cc
helenol: Truncating divide vs. modulo.
Let's say `rows * cols = 5` and `subsample_factor = 3`.
`5/3 = 1` in truncating division land.
But the check `i % 3 == 0` will return 0 twice between `0 < i < 5`.

voxblox_ros/src/intensity_server.cc
helenol: I dunno I think it's pretty easy? Rotates + normalizes one vector, that's just the offset from the image center. What's the intermediate one? unnormalized_bearing_vector? unrotated_bearing_vector?

helenol: added a comment

voxblox_ros/include/voxblox_ros/tsdf_server.h
mfehr: you are right of course, must have mixed it up with the other member that actually stores the mesh

PR #195, opened 2018-08-17
Fix up planning utils to make sense and be useful for planning.

helenol:
Basically refactors the sphere look-ups to happen in planning utils, instead of spread out through the code.
Also fixes the sphere setting in ESDFs to make some physical sense and give bounded values.

Also adds LongIndexVector.

### DIFF
--- a/voxblox/include/voxblox/utils/planning_utils.h
+++ b/voxblox/include/voxblox/utils/planning_utils.h
@@ -1,8 +1,6 @@
 #ifndef VOXBLOX_UTILS_PLANNING_UTILS_H_
 #define VOXBLOX_UTILS_PLANNING_UTILS_H_
 
-#include <algorithm>
-
 #include "voxblox/core/layer.h"
 #include "voxblox/core/voxel.h"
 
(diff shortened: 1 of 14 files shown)


voxblox/src/core/esdf_map.cc
```diff
@@ -4,7 +4,7 @@ namespace voxblox {
 
 bool EsdfMap::getDistanceAtPosition(const Eigen::Vector3d& position,
                                     double* distance) const {
-  constexpr bool interpolate = false;
+  constexpr bool interpolate = true;
```
ZacharyTaylor: should this be exposed or do we always want to interpolate?

voxblox/test/test_voxbloxpy.py
```diff
@@ -4,34 +4,36 @@
 
 import timeit
 
-import sys;
-sys.path.append('/Users/mereweth/snappy/tsdf_catkin_ws/devel/.private/voxblox/lib/')
+import sys
+sys.path.append(
+    '/Users/mereweth/snappy/tsdf_catkin_ws/devel/.private/voxblox/lib/')
```
ZacharyTaylor: won't these tests fail for anyone without these hardcoded paths?

voxblox_ros/src/esdf_server.cc
```diff
@@ -112,12 +114,10 @@ void EsdfServer::updateMesh() {
     publishTraversable();
   }
 
-  if (publish_esdf_map_ && esdf_map_pub_.getNumSubscribers() > 0) {
-    const bool only_updated = false;
-    voxblox_msgs::Layer layer_msg;
-    serializeLayerAsMsg<EsdfVoxel>(esdf_map_->getEsdfLayer(), only_updated,
-                                   &layer_msg);
-    esdf_map_pub_.publish(layer_msg);
+  // VERY important to call this before calling the mesh update, as this clears
```
ZacharyTaylor: Should these be seperate flags then?

voxblox/include/voxblox/utils/planning_utils.h
```diff
@@ -1,119 +1,161 @@
...
-                                   const Point& center,
-                                   const FloatingPoint radius) {
-  CHECK_NOTNULL(layer);
-  // search a cube with side length 2*radius
-  const FloatingPoint voxel_size = layer->voxel_size();
-  for (FloatingPoint x = -radius; x <= radius; x += voxel_size) {
-    for (FloatingPoint y = -radius; y <= radius; y += voxel_size) {
-      for (FloatingPoint z = -radius; z <= radius; z += voxel_size) {
+// Gets the indices of all points within the sphere.
+template <typename VoxelType>
```
ZacharyTaylor: does it make sense to break utils into planning_utils.h and planning_utils_inl.h if you are needing to foward declare some functions?

voxblox_ros/src/esdf_server.cc
helenol: Aah this is a left-over comment... I abandoned the incremental esdf update stuff for now (was getting to be too big of a mess, might require a bigger refactor). It actually *is* different flags, since esdf blocks have their own "updated" field. It's already a bit of a mess with the TSDF block updated field, since both the esdf update and the mesh update share the one flag. It might make sense to do a bitmask or something to be honest. But yeah, I'll remove this comment and we should discuss how to best handle this.

voxblox/include/voxblox/utils/planning_utils.h
helenol: Yep.

voxblox/test/test_voxbloxpy.py
helenol: Yeah... Maybe we should discuss moving/taking out the python bindings and tango interface, especially since we're not actually maintaining it. @mfehr @ZacharyTaylor

voxblox/src/core/esdf_map.cc
helenol: I'll expose it, especially because of the performance difference you mentioned...

PR #198, opened 2018-08-20
Feature/icp

ZacharyTaylor:
First go at voxblox ICP, taken from old IROS drone race code and cleaned up for voxblox and TSDFs.

Not sure about if we need the PCL header guards, but the `getTransformFromCorrelation` and `getTransformFromMatchedPoints` while undergoing significant cosmetic changes were taken from their implementation.

Also the speed of ICP depends on a lot of params, espically the subsampling but is usually the same order of mag as the fast integrator. Main bottleneck seems to be the gradient calc.

For tunnels dataset:
SM Timing
-----------
icp                         292     04.227528       (00.014478 +- 00.003116)        [00.001763,00.023458]
integrate/fast         292     07.782786       (00.026653 +- 00.002732)        [00.009848,00.081340]

Seems to work fairly well on tunnels dataset
mesh


voxblox/src/alignment/icp.cc
```diff
@@ -0,0 +1,224 @@
...
+
+  PointsMatrix src_demean = src.colwise() - src_center;
+  PointsMatrix tgt_demean = tgt.colwise() - tgt_center;
+
+  // align
+  return getTransformFromCorrelation(src_demean, src_center, tgt_demean,
+                                     tgt_center, T);
+}
+
+void ICP::matchPoints(const Layer<TsdfVoxel> *tsdf_layer,
```
mfehr: I think this tsdf_layer could and should be a cref, wdyt?

mfehr: as a consequence, the same goes for the other functions of the ICP class

helenol:
Ok pushed a few changes...
 - Added transform and TF output to the new exciting ICP correction frame.
 - Properly keep track of transform, controlled by a parameter (can reset initial guess each time).
 - Only run ICP if `enable_icp` flag is on.

helenol:
Oh yes and also added:
 -  Timestamp tolerance sec will now check the timestamp tolerance in the Transformer. 
`<param name="timestamp_tolerance_sec" value="0.50" />`

PR #201, opened 2018-08-21
Interpolate transforms, add input pointcloud queueing, and properly support laser pointclouds

helenol:
3 main changes:
 - Adds a queue to input pointclouds if the transform for them was not yet available. The previous behavior was to just use the latest transform in such cases, but this is a *bad idea* if your pointcloud source and pose source aren't from the same device.
 - Adds interpolation to the transformer poses in case the pose is outside a small time margin from the exact pose. Again should have had this much earlier... Needed if pointcloud and pose source aren't the exact same timestamps.
 - Adds proper support for pointclouds with intensities rather than RGB inputs. Colormaps the intensities according to a few options, described in readme.

short0, stairwell and pipes mapped by intensity with ironbow colormap:
image


voxblox_ros/src/transformer.cc
```diff
@@ -143,29 +147,50 @@ bool Transformer::lookupTransformQueue(const ros::Time& timestamp,
...
+    int64_t offset_newest_ns = (it->header.stamp - timestamp).toNSec();
+    // We already checked that this is not the beginning.
+    it--;
+    Transformation T_G_D_oldest;
+    tf::transformMsgToKindr(it->transform, &T_G_D_oldest);
+    int64_t offset_oldest_ns = (timestamp - it->header.stamp).toNSec();
+
+    // Interpolate between the two transformations using the exponential map.
+    FloatingPoint t_diff_ratio = static_cast<FloatingPoint>(offset_oldest_ns) /
+                                 (offset_newest_ns + offset_oldest_ns);
```
ZacharyTaylor: a nitpick but could we also cast the denominator?

voxblox_ros/src/tsdf_server.cc
```diff
@@ -52,6 +52,30 @@ void TsdfServer::getServerConfigFromRosParam(
   } else {  // Default case is gray.
     color_mode_ = ColorMode::kGray;
   }
+
+  // Color map for intensity pointclouds.
+  std::string intensity_colormap("rainbow");
+  float intensity_max_value = 100.0;
```
ZacharyTaylor: this default value is a little buried, could it be set from a constexpr that is in the header?

voxblox_ros/src/tsdf_server.cc
```diff
@@ -187,39 +212,59 @@ void TsdfServer::processPointCloudMessageAndInsert(
...
-              pointcloud_pcl.points[i].b, pointcloud_pcl.points[i].a));
+  // Convert differently depending on RGB or I type.
+  if (color_pointcloud) {
+    pcl::PointCloud<pcl::PointXYZRGB> pointcloud_pcl;
+    // pointcloud_pcl is modified below:
+    pcl::fromROSMsg(*pointcloud_msg, pointcloud_pcl);
+    points_C.reserve(pointcloud_pcl.size());
+    colors.reserve(pointcloud_pcl.size());
+    for (size_t i = 0; i < pointcloud_pcl.points.size(); ++i) {
+      if (!std::isfinite(pointcloud_pcl.points[i].x) ||
```
ZacharyTaylor: I feel the if (color_pointcloud) could go inside the loop and save a lot of code double up

voxblox_ros/src/transformer.cc
```diff
@@ -143,29 +147,50 @@ bool Transformer::lookupTransformQueue(const ros::Time& timestamp,
     }
   }
 
+  // Match found basically means an exact match.
```
ZacharyTaylor: Should we reduce the timestamp_tolerance now that we have interpolation, make it like 1ms?

voxblox_ros/src/tsdf_server.cc
helenol: Couldn't do this because it needs to be converted to different PCL datatypes `<PointXYZRGB>`  vs `<PointXYZI>` and I couldn't figure out how to do this any nicer :(

PR #214, opened 2018-09-18
Remove voxblox tango interface and python bindings

helenol:
Both are moved over publically to: https://github.com/ethz-asl/voxblox_deprecated

The python bindings are untested and unmaintained, and now probably completely 200% broken since so much has changed in the ESDF interface.
Project tango is over and also likewise unmaintained.

If someone wants to volunteer to maintain these, please let us know, and we can discuss other solutions. In the meantime they're available in their latest state at the repo above.


ZacharyTaylor:
LGTM

PR #221, opened 2018-09-24
Add functions to compute outer bounds of map.

helenol:
To be used mostly for planning. :) Accurate down to block level.

### DIFF
--- a/voxblox/include/voxblox/utils/planning_utils.h
+++ b/voxblox/include/voxblox/utils/planning_utils.h
@@ -31,6 +31,12 @@ void clearSphereAroundPoint(const Point& center, const FloatingPoint radius,
                             const FloatingPoint max_distance_m,
                             Layer<VoxelType>* layer);
 
+// Utility function to get map bounds from an arbitrary layer.
+template <typename VoxelType>
+void computeMapBoundsFromLayer(const voxblox::Layer<VoxelType>& layer,
+                               Eigen::Vector3d* lower_bound,
+                               Eigen::Vector3d* upper_bound);
+
 }  // namespace utils
 }  // namespace voxblox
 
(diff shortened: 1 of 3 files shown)


voxblox/include/voxblox/utils/planning_utils.h
```diff
@@ -31,6 +31,12 @@ void clearSphereAroundPoint(const Point& center, const FloatingPoint radius,
                             const FloatingPoint max_distance_m,
                             Layer<VoxelType>* layer);
 
+// Utility function to get map bounds from an arbitrary layer.
+template <typename VoxelType>
+void computeMapBoundsFromLayer(const voxblox::Layer<VoxelType>& layer,
```
mfehr: maybe add a comment that it's on the block level, not anything more detailed?

voxblox/include/voxblox/utils/planning_utils_inl.h
```diff
@@ -131,6 +131,39 @@ void clearSphereAroundPoint(const Point& center, const FloatingPoint radius,
...
+  FloatingPoint block_size = layer.block_size();
+
+  BlockIndexList all_blocks;
+  layer.getAllAllocatedBlocks(&all_blocks);
+
+  Point lower_bound_pt;
+  Point upper_bound_pt;
+  bool first_block = true;
+
+  for (const voxblox::BlockIndex& block_index : all_blocks) {
```
mfehr: Just a suggestion, but since you anyways limit the accuracy of this method to the block level, why not just compute the bounds on the block indices and then just compute the metric bounds in the end? It's simpler and faster.

helenol: :+1: good call, will change!

PR #223, opened 2018-10-24
remove check

ZacharyTaylor:
The check can fail for valid meshes. The connection process may result in a triangle using the same vertex for two of its corners, in this case the triangle indices are removed but not the vertex. This can result in more vertices than indices. This situation generally only occurs on meshs that have only one triangle meshed and it is in an extreme corner.


ZacharyTaylor:
Yeah I came to a similar conclusion as any solution would involve deleting from the middle a vector.

PR #229, opened 2018-11-17
Docs

ZacharyTaylor:
Now that more people are using it I tried to find a nice solution for documenting the voxblox API.

Doxygen seems to be the only automatic documentation tool, that can parse voxblox. I had previously tried some newer options, but they couldn't handle the templates or eigen. It can be setup and used fairly simply with rosdoc_lite http://wiki.ros.org/rosdoc_lite, however this produces ugly websites with no search function and very limited options to alter things. An example of such a site is http://docs.ros.org/melodic/api/octomap/html/classoctomap_1_1KeyRay.html

In the end I went with Exhale. Its a python tool that automates:
doxygen xml -> breathe -> sphinx -> readthedocs

Which lets us customize things and also combine the API with the wiki pages. I have a test version hosted on readthedocs, but we could always change this to a github page or something. This setup has the nice feature that it will automatically update the website on changes.
Temporary homepage: https://voxblox.readthedocs.io/en/latest/#
Example API page: https://voxblox.readthedocs.io/en/latest/api/classvoxblox_1_1EsdfMap.html 

Not as big a commit as it looks, just bloats up as there was some find and replace + doxygen file + moving in the wiki pages and changing. Mainly just these files are new https://github.com/ethz-asl/voxblox/tree/docs/docs


ZacharyTaylor:
Also @mfehr I had a go at running doxygen on Maplab and I am not sure what the other guy did wrong because it just worked for me.

PR #230, opened 2018-11-20
Fix compilation on lynx xenial

mfehr:
voxblox did not build on the lynx-xenial jenkins server


mfehr:
test this please (temporarily switch to always build on lynx xenial)

PR #241, opened 2019-02-26
Convention writing of mesh_ply header to match  pcl-1.7(Kinetic)

fabioruetz:


helenol:
add to whitelist

voxblox/src/io/mesh_ply.cc
```diff
@@ -86,7 +86,7 @@ bool outputMeshAsPly(const std::string& filename, const Mesh& mesh) {
   }
   if (mesh.hasTriangles()) {
     stream << "element face " << mesh.indices.size() / 3 << std::endl;
-    stream << "property list uchar int vertex_index" << std::endl;
+    stream << "property list uchar int vertex_indices" << std::endl;  // pcl-1.7(ros::kinetic) breaks ply convention by not using "vertex_index"
```
helenol: Could you make this be controlled by a parameter/option/something? We'll need to switch back for melodic.

helenol:
Actually on second thought this would break our compatibility with any other mesh visualization other than PCL 1.7. We should probably just fix this in the parser: i.e., replace "vertex_index" with "vertex_indices" in the string stream before we load if the correct version of PCL is defined. Let's talk about this tomorrow before you merge, @fabioruetz.

mfehr:
@fabioruetz what was the conclusion of the discussion? Are these PLYs now only compatible with pcl1.7?

PR #242, opened 2019-03-05
Made Interpolator::getVoxelsAndQVector public for use in voxgraph

victorreijgwart:
This PR makes voxblox's Interpolator::getVoxelsAndQVector(...) method public.

Having access to the Q-vector entries in other packages allows them to calculate the analytic gradient of an SDF's trilinear interpolation function directly. This is for example useful in [voxgraph](https://github.com/ethz-asl/mbzirc_2020_challenge_3/tree/victorr/voxgraph_mapper/voxgraph), where we use this gradient to register a collection of SDF submaps.


mfehr:
ok to test

PR #245, opened 2019-03-08
Sorting rays for FastTsdfIntegrator

mfehr:
For large voxel sizes smaller geometry will inevitably get cleared away. https://github.com/ethz-asl/voxblox/pull/239 exposed the "sparsity compensation factor" setting, which basically increases the weight of rays inside the truncation distance with respect to the rays outside to alleviate these effects, however, it also makes it harder to update a changing map, i.e. clear objects that have moved.

This PR tries to further improve this, by adding some more logic to fix that: 

 * ~~The first thing I tried was to correct how these clearing rays behave near the surface. The basic idea is that a ray outside the truncation distance should not contribute to the weighted average of the TSDF when passing a voxel with positive SDF that has previously been observed to be inside the truncation distance. The reason is that the only constraint on the geometry that these rays should be able to enforce is that nothing is obstructing the line of sight between sensor and measurement. If the voxel value is positive, this constraint is already fulfilled and updating the voxel with the weighted average of the truncation distance and the previous value only distorts geometry. That being said, this feature doesn't solve the problem entirely, if a voxel lies just below the surface, hence has a negative SDF, a ray passing near the outside of the surface will still update the SDF of that voxel and distort the geometry.~~
 **Edit:** Even though this logic helps preserve small geometry, it leads to problems later in planning, since it makes it harder to remove dynamic objects completely. Was removed again.

 * This one is rather simple, but was quite effective. I sort the rays before integration, such that close geometry is integrated first. For the FastTsdfIntegrator this makes more sense, since it integrates the rays backwards from point to sensor origin and aborts integration for a ray if the voxels have been touched by another ray in the same scan.

## Experiments

I ran some tests to see how these changes influences the 3D reconstruction:

| Experiment                             |      |      |      |        |        |        |       |
|----------------------------------------|------|------|------|--------|--------|--------|--------|
| integrator                             | fast | fast | fast | fast   | fast   |  fast  |  fast  |
| carving ignores freespace near surface | no   | yes  | no   | yes    |  no    |  yes   |  yes   |
| weighting (sparsity compensation)      | no   | no   | 10  | 10    |  no    |  no    |  2     |
| integration order                      |  mixed | mixed  | mixed  |  mixed | sorted  |   sorted     |  sorted  |
|                                        |      |      |      |        |        |        |       |
|  img                                      | img | img | img | img |img |img |img |

**The full experiments with more data/pictures can be found [here](https://gist.github.com/mfehr/4d8dcf522cec4ccbb7b6eec553ecc149).**

## Parameter

At least based on these qualitative tests, the current best parameters are:
 * sorted rays, some down-weighting of the voxel carving rays (e.g. 2x).

The ROS parameters are:
```xml
      <param name="method" value="fast" />
      <param name="use_sparsity_compensation_factor" value="true" />
      <param name="sparsity_compensation_factor" value="2.0" />
      <param name="use_const_weight" value="true" />
      <param name="use_weight_dropoff" value="true" />
      <param name="integration_order_mode" value="sorted" />
```

### DIFF
--- a/voxblox/src/integrator/integrator_utils.cc
+++ b/voxblox/src/integrator/integrator_utils.cc
@@ -2,16 +2,39 @@
 
 namespace voxblox {
 
-// Small class that can be used by multiple threads that need mutually exclusive
-// indexes to the same array, while still covering all elements.
-// The class attempts to ensure that the points are read in an order that gives
-// good coverage over the pointcloud very quickly. This is so that the
-// integrator can be terminated before all points have been read (due to time
-// constraints) and still capture most of the geometry.
+ThreadSafeIndex* ThreadSafeIndexFactory::get(const std::string& mode,
+                                             const Pointcloud& points_C) {
+  if (mode == "mixed") {
+    return new MixedThreadSafeIndex(points_C.size());
+  } else if (mode == "sorted") {
+    return new SortedThreadSafeIndex(points_C);
+  } else {
+    LOG(FATAL) << "Unknown integration order mode: '" << mode << "'!";
+  }
+  return nullptr;
+}
+
 ThreadSafeIndex::ThreadSafeIndex(size_t number_of_points)
-    : atomic_idx_(0),
-      number_of_points_(number_of_points),
+    : atomic_idx_(0), number_of_points_(number_of_points) {}
+
+MixedThreadSafeIndex::MixedThreadSafeIndex(size_t number_of_points)
+    : ThreadSafeIndex(number_of_points),
       number_of_groups_(number_of_points / step_size_) {}
+
+SortedThreadSafeIndex::SortedThreadSafeIndex(const Pointcloud& points_C)
+    : ThreadSafeIndex(points_C.size()) {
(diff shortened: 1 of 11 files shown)


voxblox/src/integrator/integrator_utils.cc
```diff
@@ -21,117 +44,27 @@ bool ThreadSafeIndex::getNextIndex(size_t* idx) {
...
+  const size_t position_in_group = sequential_idx / number_of_groups_;
 
   return group_num * step_size_ + position_in_group;
 }
 
-// Lookup table for the order points in a group should be read in. This is
-// simply a list from 0 to 1023 where each number has had the order of its
-// bits reversed.
-// clang-format off
-const std::array<size_t, ThreadSafeIndex::step_size_>
```
helenol: Only bit I don't get is how you factored this out?

mfehr: If I remember correctly, it wasn't used anymore. It was just left over from an older version of the threadsafe index.

PR #252, opened 2019-04-02
Feature/mesh integrator config params

alexmillane:
Adds a function for retrieving mesh integrator params from the ros_param server. The justification is the same as the other functions in this file, it stops library clients having to rewrite parameter fetching code.


alexmillane:
@mfehr Good point. Done now!

PR #254, opened 2019-04-03
add extra ROS-->PCL conversion which does not have the intensity field

lsolanka:
Some pointclouds might not have the 'intensity' field which results in the conversion
functions spamming the console with warning messages. This patch checks if the 'intensity'
field exists in the PointCloud2 message. If it doesn't it uses a dummy intensity value.

I don't have the coloured or PointXYZI clouds available so I unfortunately wasn't able to check those.


helenol:
ok to test

PR #255, opened 2019-04-12
Initialize all layer attributes in deep copy constructor

victorreijgwart:
This is a very small fix that adds initialization of the `voxel_size_inv_` and `voxels_per_side_inv_` attributes to the `Layer::Layer::Layer(const Layer& other)` deep copy constructor.

Without this fix `layer.voxel_size_inv()` returns `0` for copied layers. I assume this is not intentional, but if I'm missing something - let me know.


mfehr:
ok to test

mfehr:
add to whitelist

PR #256, opened 2019-04-17
WIP: fixing incremental ESDF integration with dynamic obstacles

fabianbl:
with the incremental esdf update, we realized that the esdf map is not always clearing dynamic obstacles. however, the dynamic obstacles get cleared within the tsdf map:

Selection_017

we get the following corresponding esdf map with especially many traces of dynamic obstacles on the upper right side of the map:

Selection_016

by adding more voxels to the raise queue first, the clearing seems to work in the esdf map as well, but I did not yet exactly find out why:

Selection_015

I saw that you mentioned in the paper that in order to reduce bookkeeping, all voxels get raised first, which does not seem to be the case in the current implementation.


helenol:
Fixed in #259

PR #267, opened 2019-05-22
Split update boolean into multiple states, improve timings for mesh, esdf, and map publishing

helenol:
***DEPENDS ON https://github.com/ethz-asl/voxblox/pull/266***

Addresses a few major pain-points with how timing for various updates was done.
Previously, the ESDF (if any) would be updated and the updated maps would be sent out when the mesh update would happen. This is horribly unintuitive, and tied everything to the same timer. This was because there was only one "updated" flag for the block.
Now it's changed to a bitset, with 3 bits: mesh, map, and esdf, so that each operation can take place independently.

**NEW parameters**:
 - `publish_map_every_n_sec` -- how often to publish the map
- `publish_pointclouds_on_update` -- whether to publish latest output visualization pointclouds whenever the map is updated (i.e., new pointclouds come in), otherwise will publish with the mesh update.
 - `update_esdf_every_n_sec` -- how often to update the ESDF.

**Removed parameters**:
 - `publish_tsdf_info` -- this name made no sense (actually published output pointclouds), replaced completely with `publish_pointclouds` which was already there.


voxblox/include/voxblox/core/block.h
```diff
@@ -11,7 +12,12 @@
 
 namespace voxblox {
 
-/** An nxnxn container holding VoxelType. It is aware of its 3D position and
+namespace Update {
+/// State of which derived things still need to be updated.
+enum State { kMap, kMesh, kEsdf, kCount };
```
mfehr: Might be a bit nit-picky: The naming `Update` here makes it look like a class.You could define it as a struct. Or directly use an enum class called `UpdateState`. Also I'm not sure "state" makes sense here, at least when I first read this part I was confused what this actually is, and imagined some state-machine like thing where this enum describes what updates you have done already in sequential order. What do you think about `UpdateFlag` or `Update::Flag`?

voxblox/include/voxblox/core/block.h
```diff
@@ -210,12 +168,14 @@ class Block {
   FloatingPoint block_size() const { return block_size_; }
 
   bool has_data() const { return has_data_; }
-  bool updated() const { return updated_; }
+  std::bitset<Update::kCount> updated() const { return updated_; }
```
mfehr: is this a copy on purpose? otherwise use cref

voxblox/include/voxblox/core/layer.h
```diff
@@ -190,12 +190,12 @@ class Layer {
     }
   }
 
-  void getAllUpdatedBlocks(BlockIndexList* blocks) const {
+  void getAllUpdatedBlocks(Update::State bit, BlockIndexList* blocks) const {
```
mfehr: nit: not sure this is self-explanatory, maybe consider the renaming suggestion above, I think `void getAllUpdatedBlocksconst Update::Flag flag, ...)` would be more intuitive.

voxblox_ros/include/voxblox_ros/esdf_server.h
```diff
@@ -32,19 +32,17 @@ class EsdfServer : public TsdfServer {
...
   virtual bool saveMap(const std::string& file_path);
   virtual bool loadMap(const std::string& file_path);
 
-  /**
-   * Call updateMesh if you want everything updated; call this specifically
-   * if you don't want the mesh or visualization.
-   */
+  void updateEsdfEvent(const ros::TimerEvent& event);
+
+  /// Call this to update the ESDF based on latest state of the TSDF map.
```
mfehr: add the incremental part and that it only consideres updated parts of the TSDF map?

voxblox_ros/src/tsdf_server.cc
```diff
@@ -193,13 +134,126 @@ TsdfServer::TsdfServer(const ros::NodeHandle& nh,
...
+  }
+  color_map_->setMaxValue(intensity_max_value);
 }
 
 // Check if all coordinates in the point are finite
-template<typename Point>
-bool isPointFinite(const Point& point) {
-    return std::isfinite(point.x) && std::isfinite(point.y) &&
-           std::isfinite(point.z);
+template <typename PCLPoint>
```
mfehr: do we have a tools/utils header somewhere we can stick this conversion stuff?

docs/pages/The-Voxblox-Node.rst
```diff
@@ -231,15 +234,14 @@ Output Parameters
...
-  Rate at which the mesh topic will be published to, a value of 0 disables. Note, this will not trigger any other mesh operations, such as generating a ply file.
 ``publish_tsdf_map`` `false`
   Whether to publish the complete TSDF map periodically over ROS topics.
 ``publish_esdf_map`` `false`
   Whether to publish the complete ESDF map periodically over ROS topics.
-``publish_tsdf_info`` `false`
-  Enables publishing of ``tsdf_pointcloud``, ``surface_pointcloud`` and ``occupied_nodes``.
-``publish_pointclouds`` | If true the tsdf and esdf (if generated) is published as a pointcloud when the mesh is updated.
+``publish_pointclouds`` `false`
+  If true the tsdf and esdf (if generated) is published as a pointcloud when the mesh is updated or whenever the map is updated if `publish_pointclouds_on_update` is set to true as well.
```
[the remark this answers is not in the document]
helenol: No sorry, bad description. 2 options:
If `publish_pointclouds_on_update`: every time a new *input* pointcloud is received, re-publish visualization pointclouds.
If NOT `publish_pointclouds_on_update` but `publish_pointclouds`: just do the pointcloud publishing when the mesh is updated (since that's how often you change the visualization). I'll change the description to be clearer.

voxblox/include/voxblox/core/block.h
helenol: How about Status? Update::Status kind of makes sense to me. The namespace is just to push it so that the constants don't have global scope; i.e., have to refer to it as Update::kMesh rather than voxblox::kMesh. Can also do the same with a class enum wrapper. What do you prefer?

mfehr: sounds good, I would go for an enum class, at least naming wise that would be more consistent with the style guide.

voxblox_ros/src/tsdf_server.cc
helenol: moved to voxblox_ros/conversions.h

helenol:
FYI couldn't use an enum class (we want the OPPOSITE of hard-typed enums here since we use tehm as ints); switched to an enum struct. I think addressed rest of comments.

PR #283, opened 2019-07-17
Introduced a tsdf voxel weight and other additional options.

fabianbl:
Note: As discussed with @victorreijgwart, this PR will go into `devel/7s` in a first stage.

The changes introduced in this PR do not change the default behavior of the tsdf integration.

The following parameters were added to the `TsdfIntegratorBase::Config`:

- `voxel_weight_factor`: allows to give an integrator a specific weight on voxels. this factor will either be divided by the squared point depth, or - if constant weight is enabled - be set as the voxel weight.
- `optical_axis_convention`: by default uses z as this is the most commonly used convention. however, it is now also possible to set another axis as the default one
- `use_only_first_point_for_clearing` (only `MergedTsdfIntegrator` specific): this was already enabled, but might not be the desired behavior in all cases as the overall weight on clearing rays will be much weaker, obviously depending on the number of points

### DIFF
--- a/voxblox/include/voxblox/integrator/tsdf_integrator.h
+++ b/voxblox/include/voxblox/integrator/tsdf_integrator.h
@@ -68,9 +68,9 @@ class TsdfIntegratorBase {
     bool use_sparsity_compensation_factor = false;
     float sparsity_compensation_factor = 1.0f;
 
-    // The coordinate axis convention used for the optical axis. This is used
-    // when computing the voxel weights w.r.t. to their depth and will have no
-    // influence when using a constant voxel weight.
+    // The coordinate axis that goes forward along the optical axis of the
+    // sensor. This is used when computing the voxel weights w.r.t. to their
+    // depth and will have no influence when using a constant voxel weight.
     CoordinateAxis optical_axis_convention = CoordinateAxis::kZ;
 
     size_t integrator_threads = std::thread::hardware_concurrency();
@@ -79,8 +79,13 @@ class TsdfIntegratorBase {
     /// rays. Options: "mixed", "sorted"
     std::string integration_order_mode = "mixed";
 
-    /// merge integrator specific
+    /// Merge integrator specific.
     bool enable_anti_grazing = false;
+
+    // Merge integrator specific. If set to true, this does the following: When
+    // computing the merged point coordinate/weight/color, this only takes into
+    // account the first point of all points contained in a voxel to speed up
+    // the procress, but also this will decrease the impact of clearing rays.
     bool use_only_first_point_for_clearing = true;
 
     /// fast integrator specific
--- a/voxblox/include/voxblox/integrator/tsdf_integrator.h
+++ b/voxblox/include/voxblox/integrator/tsdf_integrator.h
@@ -71,7 +71,7 @@ class TsdfIntegratorBase {
     // The coordinate axis that goes forward along the optical axis of the
     // sensor. This is used when computing the voxel weights w.r.t. to their
     // depth and will have no influence when using a constant voxel weight.
-    CoordinateAxis optical_axis_convention = CoordinateAxis::kZ;
+    CoordinateAxis sensor_depth_direction_axis = CoordinateAxis::kZ;
 
     size_t integrator_threads = std::thread::hardware_concurrency();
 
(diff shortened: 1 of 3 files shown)


voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -56,17 +56,23 @@ class TsdfIntegratorBase {
...
     FloatingPoint min_ray_length_m = 0.1;
     FloatingPoint max_ray_length_m = 5.0;
+    float voxel_weight_factor = 1.0f;
     bool use_const_weight = false;
     bool allow_clear = true;
     bool use_weight_dropoff = true;
     bool use_sparsity_compensation_factor = false;
     float sparsity_compensation_factor = 1.0f;
 
+    // The coordinate axis convention used for the optical axis. This is used
```
dymczykm: could you elaborate on the optical axis. You mean this is the axis that goes forward along the optical axis of the sensor?

voxblox/include/voxblox/integrator/tsdf_integrator.h
```diff
@@ -75,6 +81,7 @@ class TsdfIntegratorBase {
 
     /// merge integrator specific
     bool enable_anti_grazing = false;
+    bool use_only_first_point_for_clearing = true;
```
dymczykm: add a comment what it does?

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -372,9 +386,10 @@ void MergedTsdfIntegrator::bundleRays(
 void MergedTsdfIntegrator::integrateVoxel(
     const Transformation& T_G_C, const Pointcloud& points_C,
     const Colors& colors, bool enable_anti_grazing, bool clearing_ray,
-    const std::pair<GlobalIndex, AlignedVector<size_t>>& kv,
+    const GlobalIndex& voxel_index,
+    const AlignedVector<size_t>& point_C_indices,
```
dymczykm: I'm not sure why is this aligned, but ok...

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -383,9 +398,9 @@ void MergedTsdfIntegrator::integrateVoxel(
   Point merged_point_C = Point::Zero();
   FloatingPoint merged_weight = 0.0;
 
-  for (const size_t pt_idx : kv.second) {
-    const Point& point_C = points_C[pt_idx];
-    const Color& color = colors[pt_idx];
+  for (const size_t point_idx : point_C_indices) {
+    const Point& point_C = points_C[point_idx];
```
dymczykm: what about check_lt?

voxblox/src/integrator/tsdf_integrator.cc
```diff
@@ -229,11 +230,24 @@ float TsdfIntegratorBase::computeDistance(const Point& origin,
 // Thread safe.
 float TsdfIntegratorBase::getVoxelWeight(const Point& point_C) const {
   if (config_.use_const_weight) {
-    return 1.0f;
+    return config_.voxel_weight_factor;
```
rdube: What is the `_factor` part?

rdube: Got the info from below although can we assume that the const weight is the same as the factor?

voxblox/src/integrator/tsdf_integrator.cc
fabianbl: yes, me neither, but it's used in other parts of the code as well, so probably it's better to leave it as it is for now.

voxblox/include/voxblox/integrator/tsdf_integrator.h
fabianbl: yes, exactly, I adjusted the description

voxblox/src/integrator/tsdf_integrator.cc
fabianbl: true, so to be correct, `factor` only applies to the lower part, where it is a factor to the inverse squared depth. but as we have `use_constant_weight` disabled by default, I thought that it's probably better to call it factor.

PR #288, opened 2019-07-29
Improved tsdf voxel merging and forEachVoxelInLayer(...) method.

fabianbl:
This PR contains the following main changes:

- When using the `mergeLayerAintoLayerB`, we don't check if the `has_data` flag of the blocks is set to true. As far as I could see, this doesn't make much sense though as we never set it in any of the tsdf integrators. The other option would obviously be to set this flag directly in the integrators, but then this is also hard to maintain as we would need a mechanism to detect when a block is empty again. So I suggest to skip the `has_data()` check when merging the tsdf blocks. (`grep` output for `has_data` is printed at the bottom.
- Added two `forEachVoxelInLayer(...)` methods which allow to iterate through all voxels in the corresponding layer. One of them has a mutable callback, so we can modify the content of the voxels there.
- Added a `EsdfVoxel` interpolation as otherwise, calling layer subsampling in `mergeLayerAintoLayerB` would result in undefined references.
- In the incremental esdf update, deleted tsdf blocks would not be detected, but we would want to keep the number of esdf blocks bounded. Therefore, there is now an additional step which checks if all allocated esdf blocks have a counterpart in the tsdf map, otherwise they get deleted.

grep output for `has_data`:
```
./voxblox/proto/voxblox/Block.proto:  optional bool has_data = 6;
./voxblox/src/core/tsdf_map.cc:    if (!block.has_data()) {
./voxblox/test/test_tsdf_interpolator.cc:    block_ptr->has_data() = true;
./voxblox/test/test_tsdf_interpolator.cc:        block_ptr->has_data() = true;
./voxblox/include/voxblox/utils/planning_utils_inl.h:        block_ptr->has_data() = true;
./voxblox/include/voxblox/utils/planning_utils_inl.h:        block_ptr->has_data() = true;
./voxblox/include/voxblox/core/block_inl.h:  has_data_ = proto.has_data();
./voxblox/include/voxblox/core/block_inl.h:  proto->set_has_data(has_data_);
./voxblox/include/voxblox/core/block_inl.h:  has_data() = true;
./voxblox/include/voxblox/core/block_inl.h:  size += sizeof(has_data_);
./voxblox/include/voxblox/core/block.h:      : has_data_(false),
./voxblox/include/voxblox/core/block.h:  bool has_data() const { return has_data_; }
./voxblox/include/voxblox/core/block.h:  bool& has_data() { return has_data_; }
./voxblox/include/voxblox/core/block.h:  void set_has_data(bool has_data) { has_data_ = has_data; }
./voxblox/include/voxblox/core/block.h:  bool has_data_;
./voxblox/include/voxblox/integrator/merge_integration.h:        output_block->has_data() = true;
./voxblox/include/voxblox/integrator/merge_integration.h:        block->has_data() = true;
./voxblox/include/voxblox/integrator/merge_integration.h:        block->has_data() = true;
./voxblox/include/voxblox/integrator/merge_integration.h:    if (!block->has_data()) {
./voxblox/include/voxblox/test/layer_test_utils.h:        block->has_data() = true;
```

### DIFF
--- a/voxblox/include/voxblox/core/block_inl.h
+++ b/voxblox/include/voxblox/core/block_inl.h
@@ -96,10 +96,6 @@ void Block<VoxelType>::getProto(BlockProto* proto) const {
   proto->set_origin_y(origin_.y());
   proto->set_origin_z(origin_.z());
 
-  // This member does not exist within the block class definition anymore, but
-  // it will remain in the protobuf definition to maintain compatibility.
-  proto->set_has_data(true);
-
   std::vector<uint32_t> data;
   serializeToIntegers(&data);
   // Not quite actually a word since we're in a 64-bit age now, but whatever.
--- a/voxblox/include/voxblox/core/block_inl.h
+++ b/voxblox/include/voxblox/core/block_inl.h
@@ -112,15 +112,19 @@ template <typename VoxelType>
 void Block<VoxelType>::mergeBlock(const Block<VoxelType>& other_block) {
   CHECK_EQ(other_block.voxel_size(), voxel_size());
   CHECK_EQ(other_block.voxels_per_side(), voxels_per_side());
+  // LOG(INFO) << "before has_data(...) check.";
 
-  if (!other_block.has_data()) {
+  // if (!other_block.has_data()) {
+  if (false) {
     return;
   } else {
     has_data() = true;
     updated().set();
 
+    // LOG(INFO) << "before mergeVoxelAIntoVoxelB";
     for (IndexElement voxel_idx = 0;
          voxel_idx < static_cast<IndexElement>(num_voxels()); ++voxel_idx) {
+      // LOG(INFO) << "in mergeVoxelAIntoVoxelB";
       mergeVoxelAIntoVoxelB<VoxelType>(
           other_block.getVoxelByLinearIndex(voxel_idx),
           &(getVoxelByLinearIndex(voxel_idx)));
--- a/voxblox/include/voxblox/core/block_inl.h
+++ b/voxblox/include/voxblox/core/block_inl.h
@@ -112,23 +112,14 @@ template <typename VoxelType>
 void Block<VoxelType>::mergeBlock(const Block<VoxelType>& other_block) {
   CHECK_EQ(other_block.voxel_size(), voxel_size());
   CHECK_EQ(other_block.voxels_per_side(), voxels_per_side());
-  // LOG(INFO) << "before has_data(...) check.";
+  has_data() = true;
+  updated().set();
 
-  // if (!other_block.has_data()) {
-  if (false) {
-    return;
-  } else {
-    has_data() = true;
-    updated().set();
-
-    // LOG(INFO) << "before mergeVoxelAIntoVoxelB";
-    for (IndexElement voxel_idx = 0;
-         voxel_idx < static_cast<IndexElement>(num_voxels()); ++voxel_idx) {
-      // LOG(INFO) << "in mergeVoxelAIntoVoxelB";
-      mergeVoxelAIntoVoxelB<VoxelType>(
-          other_block.getVoxelByLinearIndex(voxel_idx),
-          &(getVoxelByLinearIndex(voxel_idx)));
-    }
+  for (IndexElement voxel_idx = 0;
+       voxel_idx < static_cast<IndexElement>(num_voxels()); ++voxel_idx) {
+    mergeVoxelAIntoVoxelB<VoxelType>(
+        other_block.getVoxelByLinearIndex(voxel_idx),
+        &(getVoxelByLinearIndex(voxel_idx)));
   }
 }
 
--- a/voxblox/include/voxblox/core/block_inl.h
+++ b/voxblox/include/voxblox/core/block_inl.h
@@ -74,8 +74,6 @@ template <typename VoxelType>
 Block<VoxelType>::Block(const BlockProto& proto)
     : Block(proto.voxels_per_side(), proto.voxel_size(),
             Point(proto.origin_x(), proto.origin_y(), proto.origin_z())) {
-  has_data_ = proto.has_data();
-
   // Convert the data into a vector of integers.
   std::vector<uint32_t> data;
   data.reserve(proto.voxel_data_size());
@@ -98,7 +96,9 @@ void Block<VoxelType>::getProto(BlockProto* proto) const {
   proto->set_origin_y(origin_.y());
   proto->set_origin_z(origin_.z());
 
-  proto->set_has_data(has_data_);
+  // This member does not exist within the block class definition anymore, but
+  // it will remain in the protobuf definition to maintain compatibility.
+  proto->set_has_data(true);
 
   std::vector<uint32_t> data;
   serializeToIntegers(&data);
@@ -112,7 +112,6 @@ template <typename VoxelType>
 void Block<VoxelType>::mergeBlock(const Block<VoxelType>& other_block) {
   CHECK_EQ(other_block.voxel_size(), voxel_size());
   CHECK_EQ(other_block.voxels_per_side(), voxels_per_side());
-  has_data() = true;
   updated().set();
 
   for (IndexElement voxel_idx = 0;
@@ -135,7 +134,6 @@ size_t Block<VoxelType>::getMemorySize() const {
   size += sizeof(voxel_size_inv_);
   size += sizeof(block_size_);
 
-  size += sizeof(has_data_);
   size += sizeof(updated_);
 
   if (num_voxels_ > 0u) {
--- a/voxblox/include/voxblox/core/layer.h
+++ b/voxblox/include/voxblox/core/layer.h
@@ -30,12 +30,14 @@ class Layer {
   using BlockHashMap =
       typename AnyIndexHashMapType<typename BlockType::Ptr>::type;
   using BlockMapPair = typename std::pair<BlockIndex, typename BlockType::Ptr>;
-  using VoxelAction = std::function<void(const BlockIndex& /*block_index*/,
-                                         size_t /*linear_voxel_index*/,
-                                         const VoxelType& /*voxel*/)>;
+
+  // TODO(fabianbl): Output parameters have to change to pointers.
+  using VoxelAction = std::function<void(
+      const BlockIndex& /*block_index*/, const BlockType& /*block*/,
+      size_t /*linear_voxel_index*/, const VoxelType& /*voxel*/)>;
   using MutableVoxelAction = std::function<void(
-      const BlockIndex& /*block_index*/, size_t /*linear_voxel_index*/,
-      VoxelType& /*voxel*/)>;
+      const BlockIndex& /*block_index*/, BlockType* /*block*/,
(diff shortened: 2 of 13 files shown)


voxblox/src/utils/voxel_utils.cc
```diff
@@ -18,6 +28,12 @@ void mergeVoxelAIntoVoxelB(const TsdfVoxel& voxel_A, TsdfVoxel* voxel_B) {
                                            voxel_B->color, voxel_B->weight);
 
     voxel_B->weight = combined_weight;
+  } else {
+    // If the combined voxel weight is smaller or equal to 0, and one of voxel
+    // weights is smaller than 0, this means that we try to substract the two
+    // (e.g. an integrated point cloud that we want remove again). In this
```
rdube: that we want (to) remove ..

voxblox/include/voxblox/core/layer.h
```diff
@@ -24,11 +25,18 @@ class Layer {
...
+  using BlockHashMap =
+      typename AnyIndexHashMapType<typename BlockType::Ptr>::type;
+  using BlockMapPair = typename std::pair<BlockIndex, typename BlockType::Ptr>;
+
+  using VoxelAction = std::function<void(
+      const BlockIndex& /*block_index*/, const BlockType& /*block*/,
+      const size_t /*linear_voxel_index*/, const VoxelType& /*voxel*/)>;
+  using MutableVoxelAction = std::function<void(
+      const BlockIndex& /*block_index*/, BlockType* /*block*/,
+      const size_t /*linear_voxel_index*/, VoxelType* /*voxel*/)>;
```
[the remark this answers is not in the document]
fabianbl: true, the pointers should go last

voxblox/include/voxblox/core/block_inl.h
```diff
@@ -112,19 +112,14 @@ template <typename VoxelType>
...
-    updated().set();
-
-    for (IndexElement voxel_idx = 0;
-         voxel_idx < static_cast<IndexElement>(num_voxels()); ++voxel_idx) {
-      mergeVoxelAIntoVoxelB<VoxelType>(
-          other_block.getVoxelByLinearIndex(voxel_idx),
-          &(getVoxelByLinearIndex(voxel_idx)));
-    }
+  has_data() = true;
+  updated().set();
```
Eeppii: test comment

voxblox/include/voxblox/core/layer_inl.h
```diff
@@ -39,6 +39,36 @@ Layer<VoxelType>::Layer(const LayerProto& proto)
...
+  }
+}
+
+template <typename VoxelType>
+void Layer<VoxelType>::forEachVoxelInLayer(
+    const MutableVoxelAction& voxel_action) {
+  voxblox::BlockIndexList all_blocks;
+  getAllAllocatedBlocks(&all_blocks);
+  for (const voxblox::BlockIndex& block_index : all_blocks) {
+    BlockType& block = *CHECK_NOTNULL(getBlockPtrByIndex(block_index));
```
Eeppii: Could you use
```
BlockType& block = getBlockByIndex(block_index);
```
as above (just without `const`)?

...`getBlockByIndex` will end in FATAL if the block does not exist, covering your `CHECK_NOTNULL`.

voxblox/src/integrator/esdf_integrator.cc
```diff
@@ -119,6 +119,16 @@ void EsdfIntegrator::updateFromTsdfLayer(bool clear_updated_flag) {
...
     }
   }
+
+  // Clear all esdf blocks which are no longer allocated in the corresponding
+  // tsdf layer.
+  BlockIndexList allocated_esdf_blocks;
+  esdf_layer_->getAllAllocatedBlocks(&allocated_esdf_blocks);
+  for (const BlockIndex& esdf_block_index : allocated_esdf_blocks) {
+    if (!tsdf_layer_->hasBlock(esdf_block_index)) {
+      esdf_layer_->removeBlock(esdf_block_index);
```
Eeppii: Don't know voxblox well enough - I assume the `esdf_block_index` are global, so removing one block does not shift the indices of the remaining ones? (Otherwise, the `removeBlock` in the loop could remove the wrong blocks)

voxblox/src/utils/voxel_utils.cc
```diff
@@ -18,6 +28,12 @@ void mergeVoxelAIntoVoxelB(const TsdfVoxel& voxel_A, TsdfVoxel* voxel_B) {
                                            voxel_B->color, voxel_B->weight);
 
     voxel_B->weight = combined_weight;
+  } else {
+    // If the combined voxel weight is smaller or equal to 0, and one of voxel
```
Eeppii: In the comment you mention that "...if the combined voxel weight is smaller or equal to 0 **and** one of he voxel weights is smaller than 0..." but you do not check for the latter statement.

I guess that if the combined voxel weight is below zero then anyway one of the voxel weights must be negative. Maybe adjust this in the comment (this and the suggestion below).

```suggestion
    // If the combined voxel weight is smaller or equal to 0, this means that one of voxel
```

voxblox/src/utils/voxel_utils.cc
```diff
@@ -18,6 +28,12 @@ void mergeVoxelAIntoVoxelB(const TsdfVoxel& voxel_A, TsdfVoxel* voxel_B) {
...
+    // weights is smaller than 0, this means that we try to substract the two
```
Eeppii:
```suggestion
    // weights is smaller than 0, and that we try to substract the two
```

voxblox/include/voxblox/core/layer.h
```diff
@@ -24,11 +25,18 @@ class Layer {
...
-  typedef std::shared_ptr<Layer> Ptr;
-  typedef Block<VoxelType> BlockType;
-  typedef
-      typename AnyIndexHashMapType<typename BlockType::Ptr>::type BlockHashMap;
-  typedef typename std::pair<BlockIndex, typename BlockType::Ptr> BlockMapPair;
+  using Ptr = std::shared_ptr<Layer>;
+  using BlockType = Block<VoxelType>;
+  using BlockHashMap =
+      typename AnyIndexHashMapType<typename BlockType::Ptr>::type;
+  using BlockMapPair = typename std::pair<BlockIndex, typename BlockType::Ptr>;
```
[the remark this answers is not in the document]
fabianbl: I think from the context, it should be descriptive enough

voxblox/include/voxblox/core/layer_inl.h
```diff
@@ -39,6 +39,36 @@ Layer<VoxelType>::Layer(const LayerProto& proto)
...
+void Layer<VoxelType>::forEachVoxelInLayer(
+    const VoxelAction& voxel_action) const {
+  voxblox::BlockIndexList all_blocks;
+  getAllAllocatedBlocks(&all_blocks);
+  for (const voxblox::BlockIndex& block_index : all_blocks) {
+    const BlockType& block = getBlockByIndex(block_index);
+    for (size_t linear_voxel_index = 0u;
+         linear_voxel_index < block.num_voxels(); ++linear_voxel_index) {
+      const VoxelType& voxel = block.getVoxelByLinearIndex(linear_voxel_index);
+      voxel_action(block_index, block, linear_voxel_index, voxel);
```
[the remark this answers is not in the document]
fabianbl: yes exactly!

voxblox/include/voxblox/core/layer_inl.h
```diff
@@ -39,6 +39,36 @@ Layer<VoxelType>::Layer(const LayerProto& proto)
   voxels_per_side_inv_ = 1.0f / static_cast<FloatingPoint>(voxels_per_side_);
 }
 
+template <typename VoxelType>
+void Layer<VoxelType>::forEachVoxelInLayer(
+    const VoxelAction& voxel_action) const {
+  voxblox::BlockIndexList all_blocks;
+  getAllAllocatedBlocks(&all_blocks);
+  for (const voxblox::BlockIndex& block_index : all_blocks) {
```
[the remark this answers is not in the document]
fabianbl: yes, it's a `Eigen::Vector3i`

voxblox/include/voxblox/interpolator/interpolator_inl.h
```diff
@@ -437,6 +437,15 @@ inline TsdfVoxel Interpolator<TsdfVoxel>::interpVoxel(
   return voxel;
 }
 
+template <>
+inline EsdfVoxel Interpolator<EsdfVoxel>::interpVoxel(
+    const InterpVector& q_vector, const EsdfVoxel** voxels) {
```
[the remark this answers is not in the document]
fabianbl: because the EsdfVoxel double pointer is pointing to an array of voxels

voxblox/include/voxblox/core/layer.h
```diff
@@ -2,6 +2,7 @@
 #define VOXBLOX_CORE_LAYER_H_
 
 #include <glog/logging.h>
+#include <functional>
```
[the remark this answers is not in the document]
fabianbl: true, I changed that

voxblox/src/integrator/esdf_integrator.cc
fabianbl: yes, exactly, the block indices are global

voxblox/src/utils/voxel_utils.cc
fabianbl: good point, thanks for the suggestion! yes, that's what I wanted to say, I'll change the formulation

voxblox/include/voxblox/core/block_inl.h
```diff
@@ -112,19 +112,14 @@ template <typename VoxelType>
...
-    has_data() = true;
-    updated().set();
-
-    for (IndexElement voxel_idx = 0;
-         voxel_idx < static_cast<IndexElement>(num_voxels()); ++voxel_idx) {
-      mergeVoxelAIntoVoxelB<VoxelType>(
-          other_block.getVoxelByLinearIndex(voxel_idx),
-          &(getVoxelByLinearIndex(voxel_idx)));
-    }
+  has_data() = true;
```
dymczykm: how do you know it has data?

fabianbl: @dymczykm not really, you're right. however, this `has_data` member does not seem to be consistently used through voxblox, and I would be in favor of removing it completely as quite some bookkeeping would be necessary to set it properly in all cases (let's maybe see what @victorreijgwart point is here).

wdyt?

voxblox/include/voxblox/core/block_inl.h
```diff
@@ -98,7 +96,9 @@ void Block<VoxelType>::getProto(BlockProto* proto) const {
   proto->set_origin_y(origin_.y());
   proto->set_origin_z(origin_.z());
 
-  proto->set_has_data(has_data_);
+  // This member does not exist within the block class definition anymore, but
+  // it will remain in the protobuf definition to maintain compatibility.
+  proto->set_has_data(true);
```
dymczykm: do you need to set it if it's not used?

fabianbl: probably not, will remove it

PR #294, opened 2019-09-02
Added required dependancies for fresh install

nkhedekar:


helenol:
Re: your e-mail, I'm unclear on why these are required as they are python-only libraries and not used anywhere in voxblox. This is very weird to me.
I've merged PR #299 which adds the protobuf dependencies which it looks like you were actually missing?

helenol:
@alexmillane +1 to that. :) Let's then skip the python deps (maybe we should update docs to just build voxblox_ros and voxblox_rviz_plugin?).

alexmillane:
The docs have now changed to only build voxblox_ros (which depends on voxblox_rviz_plugin). I'm closing this now. Thanks @nkhedekar for bringing this to our attention.

PR #300, opened 2019-10-16
Doc bug fix

oleg-alexandrov:
Fix a bug in the doc. Add a blurb directing the user to more


helenol:
add to whitelist (though this is just docs so we can merge directly)

PR #302, opened 2019-10-23
Clarify that the voxtel size is in meters

oleg-alexandrov:


oleg-alexandrov:
Actually I was secretly hoping that the voxtel size was not in meters but in inches. :)

PR #310, opened 2020-01-14
include gflags header

Schmluk:
Added gflags include statement to nodes, otherwise voxblox_ros won't compile on a clean system setup with 
`error: ‘ParseCommandLineFlags’ is not a member of ‘google’`


mfehr:
ok to test

PR #313, opened 2020-02-10
Force voxblox to use protobuf_catkin

helenol:
**DO NOT MERGE UNTIL THIS IS MERGED: https://github.com/ethz-asl/protobuf_catkin/pull/35**

### YOU WILL NEED TO DO A CLEAN REBUILD OF VOXBLOX AFTER THIS!!

Since we're swapping protobuf catkin to default to system library, we can swap voxblox to depending on protobuf_catkin. Main advantages this gives us is cross-compatibility with other stuff using protobuf_catkin, access to the `PROTOBUF_CATKIN_GENERATE_CPP2` function which puts protos *not* in the build folder, and a clean-up of all dependent packages needing to link against protobuf over and over again.

The .rosinstall files have been updated to reflect this change.

Any packages that depend on this can ditch the extra proto linking step as a result of this change.

Let me know if there's any issues.


voxblox/include/voxblox/core/block.h
```diff
@@ -7,7 +7,7 @@
 #include <memory>
 #include <vector>
 
-#include "./Block.pb.h"
+#include "voxblox/Block.pb.h"
```
alexmillane: Thank god this is gone.

mfehr:
test this please

alexmillane:
We waiting on anything here @helenol @mfehr ?

I'm actively using this branch if that helps. Seems legit.

PR #314, opened 2020-02-24
Accessor function for voxel weight in TSDF map

gasserl:
Added a function to calculate and access the weight in the TSDF map, analog to the distance value. 

Changes:
- Added getWeightAtPosition for TSDF maps
- Added interpolator attribute to the TSDF map, used for computing the weight at any position
- Added interpolation functions getWeight, getInterpWeight and getNearestWeight to Interpolator class for voxel weights


alexmillane:
Adding an accessor for interpolated voxel weight is good. One change that I think is required is to make these function template specializations of TsdfVoxel (otherwise they won't compile for wight-less voxel types, I presume).

gasserl:
@alexmillane 
_> In my opinion including an interpolator in TsdfMap is a bit of a big change. Is there an issue with having an integrator in your code operating on the layer held by the map in question?_
The integrator itself has a layer member, which would result in a circular dependency. The ESDF map similarly also contains an interpolator member, I assume for the same reasons.

_> Adding an accessor for interpolated voxel weight is good. One change that I think is required is to make these function template specializations of TsdfVoxel (otherwise they won't compile for wight-less voxel types, I presume)._
The new functions themselves only rely on functions already specialized for TSDF and ESDF voxels. I've added unspecialized implementations of these to ensure compilation. They will report the voxel as unobserved and return a weight of zero.

mfehr:
ok to test

voxblox_ros/include/voxblox_ros/tsdf_server.h
```diff
@@ -95,7 +95,7 @@ class TsdfServer {
   void updateMeshEvent(const ros::TimerEvent& event);
   void publishMapEvent(const ros::TimerEvent& event);
 
-  std::shared_ptr<TsdfMap> getTsdfMapPtr() { return tsdf_map_; }
+  std::shared_ptr<TsdfMap> getTsdfMapPtr() const { return tsdf_map_; }
```
mfehr, line 98: if you need const access to the tsdf layer, wouldn't it make more sense to add a function to get a const reference? I think that would be the correct way to do it, or at the very least this should return a const shared ptr. Basically similar [how it is here in the layer to get blocks.](https://github.com/ethz-asl/voxblox/blob/master/voxblox/include/voxblox/core/layer.h#L104)

PR #323, opened 2020-04-21
Fix style by running clang-format once and updating include order

ntonci:
There is no functionality change in here, just running clang-format over everything once.
clang-format file is updated to preserve the order of the include blocks.
Include order has been updated such that it is consistent over the whole repo.

### DIFF
--- a/voxblox/include/voxblox/core/tsdf_map.h
+++ b/voxblox/include/voxblox/core/tsdf_map.h
@@ -1,10 +1,11 @@
 #ifndef VOXBLOX_CORE_TSDF_MAP_H_
 #define VOXBLOX_CORE_TSDF_MAP_H_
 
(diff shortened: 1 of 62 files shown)


voxblox/include/voxblox/core/tsdf_map.h
```diff
@@ -91,8 +93,8 @@ class TsdfMap {
 
   bool getWeightAtPosition(const Eigen::Vector3d& position,
                            double* weight) const;
-  bool getWeightAtPosition(const Eigen::Vector3d& position,
-                           bool interpolate, double* weight) const;
+  bool getWeightAtPosition(const Eigen::Vector3d& position, bool interpolate,
```
alexmillane, line 96: `const bool interpolate`

PR #324, opened 2020-04-29
Added key dependencies to rosinstall

mfinean:
Both numpy_eigen and catkin_boost_python_buildtool are required


mfehr:
https://github.com/ethz-asl/voxblox/pull/294/files

PR #327, opened 2020-05-26
added multi-mesh plugin and material for transparency

Schmluk:
Added support for multiple meshes and transparency for mesh visualization. Thought this feature might also be useful for others. Changes in particular include:

- Added a MultiMesh message that simply wraps the normal voxblox_msgs::Mesh for multiple publishers. This also ensures compatibility with earlier versions since the original definitions remain untouched.
- Added a multi_mesh_rviz plugin (which is almost identical to the mesh_rviz plugin) that supports multiple mesh instances in the same topic, similar to visualization_msgs/marker IDs.
- Added an ogre material based on 'BaseWhiteNoLighting' that allows to set alpha properly. Although this functionality is also available for the standard voxblox_msgs::Mesh it is not used there since the message does not contain an alpha channel (could be added if requested though). 
- Tested with two independent mesh layers as in the image below, which did not work using only one voxblox_msgs::Mesh topic, and where one of the layers is partially transparent.
- Added some quality of life improvements for both plugins:
  - Can now properly show/hide the meshes without discarding everything.
  - Can change the fixed frame after mapping is finished.
  - Can still reset the mesh via 'Reset Mesh' property, this should also reduce accidental clears.

image:
voxblox_meshing


victorreijgwart:
ok to test

Schmluk:
Edit: Added quality of life improvements for both plugins (see bottom bulletpoint of original PR). Tested both plugins.

Schmluk:
Edit2: Added longer queue for the multi mesh plugin (as it was unreliable for many meshes) and a property to reset the mesh via a single click, s.t. all previous functionality is preserved. This should be the final final version (:

PR #332, opened 2020-07-23
Added getter for const ptr.

alexmillane:
An analgoue to the function available for tsdf maps here:

https://github.com/ethz-asl/voxblox/blob/master/voxblox/include/voxblox/core/tsdf_map.h#L65


victorreijgwart (approved):
LGTM

PR #334, opened 2020-07-28
Feature/read from general stream

alexmillane:
I needed to read protobuf messages from a stream of data in memory (as opposed to in a file). I swapped loading function inputs from `fstream*` to `istream*`. They were being implicitly cast to those types anyway, when passed to protobuf functions. 

I also corrected a naming typo... This is a breaking change for anyone depending on this function... But it was quite confusing.

on top of #332

### DIFF
--- a/voxblox/include/voxblox/utils/protobuf_utils.h
+++ b/voxblox/include/voxblox/utils/protobuf_utils.h
@@ -2,6 +2,7 @@
 #define VOXBLOX_UTILS_PROTOBUF_UTILS_H_
 
 #include <fstream>
+#include <istream>
 
 #include <glog/logging.h>
 #include <google/protobuf/message.h>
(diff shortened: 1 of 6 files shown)


voxblox/include/voxblox/utils/protobuf_utils.h
```diff
@@ -10,13 +11,14 @@
 namespace voxblox {
 
 namespace utils {
-bool readProtoMsgCountToStream(std::fstream* stream_in, uint32_t* message_count,
-                               uint64_t* byte_offset);
+bool readProtoMsgCountFromStream(std::istream* stream_in,
```
alexmillane, line 14: This is naming correction that will break depending packages if they've implemented their own loading functions. 

Obviously this function is reading _from_ a stream, rather than _to_ a stream.

helenol: I think you should be able to rename this. AFAIK this should only be used deep in the internals of this stuff (note to self for future designs: hide functions like these :) )

helenol (approved):
LGTM, thanks!

voxblox/include/voxblox/utils/protobuf_utils.h
alexmillane: People implementing new layers, submap types etc have to call this function, which does seem a bit odd to me, as we've discussed before :). Imma merge and let's see if anyone complains.

PR #336, opened 2020-08-27
Add dedicated color maps for unique IDs

victorreijgwart:
This PR implements new color maps to make it easier to visualize unique IDs, with the following goals:
- knowledge of the max ID should not be required, such that new IDs can be added without ever having to change the colors of past IDs
- the colors for different IDs should remain as distinguishable as possible, so colors should not repeat and spread well across the color wheel
- the colors should communicate which IDs are subsequent, for example by being (roughly) constantly spaced on the color wheel

Two color maps are implemented, based on
1. Division by an irrational number followed, which results in a fractional left over that never repeats. See the left plot in the gif below.
2. Constant steps that are offset such that the colors on subsequent revolutions always fall in between those of previous revolutions. See the right plot in the gif below.

Illustration:
Note that the colors in question are shown in the bottom row. The colors in the top two rows only illustrate which samples are subsequent.
ezgif com-video-to-gif (2)

### DIFF
--- a/voxblox/include/voxblox/utils/color_maps.h
+++ b/voxblox/include/voxblox/utils/color_maps.h
@@ -110,7 +110,7 @@ class IrrationalIdColorMap : IdColorMap {
 
   void setIrrationalBase(float value) { irrational_base_ = value; }
 
-  virtual Color colorLookup(size_t value) const {
+  virtual Color colorLookup(const size_t value) const {
     const float normalized_color = std::fmod(value / irrational_base_, 1.f);
     return rainbowColorMap(normalized_color);
   }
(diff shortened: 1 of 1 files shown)


voxblox/include/voxblox/utils/color_maps.h
```diff
@@ -91,6 +91,72 @@ class IronbowColorMap : public ColorMap {
...
+ * An important advantage of this color map is that the colors are independent
+ * of the max ID value, e.g. previous colors don't change when adding new IDs.
+ */
+class ExponentialOffsetColorMap : IdColorMap {
+ public:
+  ExponentialOffsetColorMap() : items_per_revolution_(10u) {}
+
+  void setItemsPerRevolution(uint value) { items_per_revolution_ = value; }
+
+  virtual Color colorLookup(size_t value) const {
```
alexmillane, line 133: Here and above, I personally make pass-by-value arguments const. It's saved me a few times from making assignment-instead-of-comparison mistakes.

alexmillane (approved):
LGTM. Thanks for doing this.

voxblox/include/voxblox/utils/color_maps.h
victorreijgwart: Good call, thanks.

PR #338, opened 2020-08-31
Make names of classes derived from IdColorMap consistent

victorreijgwart:
@Schmluk thanks for pointing this out


Schmluk (approved):
LGTM

PR #348, opened 2020-11-04
make publish_pointclouds param grip for external map updates

Schmluk:
Minor fix to prevent pointclouds from being published all the time if they are disabled, draining a lot of compute.


victorreijgwart:
Ok to test

victorreijgwart (approved):
LGTM

PR #358, opened 2021-02-28
Add distance getters to tsdf map

lgulich:
Allows to directly access the distance in a TSDF map without first computing the ESDF map. This is useful for ground robots which are only interested in the environment near the surfaces.


lgulich:
test this please

PR #377, opened 2021-10-15
Add missing getter

nachovizzo:
If the member is protected it should be a way of accessing it at least as
a const reference. The other option would be to create a wrapper class
that inherits from TSDFIntegrator, and provide the missing getter... but
this is an overkill.


nachovizzo:
Closing this due inactivity

